PR #6, opened 2017-07-10
Add babel-plugin-normalize-requires

izaera:


izaera:
@jbalsas I have tested that case:

`require('./a-module.js/')`

from node.js CLI and it is illegal. It seems that you can only add a trailing `/` for folders. (And I say `it seems` because I haven't been capable of finding any documentation where it is stated what is correct and what is illegal :-( )

PR #9, opened 2017-07-11
Add babel-plugin-wrap-modules-amd

izaera:


izaera:
Sadly, it is. That’s what the plugin does.

If you have AMD modules in your node_modules (like jQuery, for example) you should not add this plugin to the build. 

In general it’s not easy to know if something is an AMD module or not. For instance, jQuery provides UMD modules and the define() call is done near the end of the module and nested inside an if. 

We could look for those kind of constructs in the plugin, but it would lead to a much more complex code and far slower, so I preferred to handle that cases by configuration. In the end, it’s conceptually the same as if you added any other incorrect Babel plugin to the build: it would break everything, too ;-).

This approach would fail if you mixed AMD and non-AMD modules in the same package but that sound like nonsense, doesn’t it?

jbalsas:
I'm merging for now then... this does what it does :D

yuchi:
> This approach would fail if you mixed AMD and non-AMD modules in the same package but that sound like nonsense, doesn’t it?

It’s nonsense if you look at a single package, it is not if you look at dependencies.

Are we asking developers to actually know the format the modules they require from packages are written in? And to be sure all of the conform to the same format? :\

izaera:
@yuchi That's true: you may have mixed dependencies and you would have to configure the build to handle it (at least for now) :-(.

We can try to find a way to make a smart plugin (not this one, but a different one) that detects if the wrapping is necessary but, as I said, the detection would not be 100% safe because `define` is not a reserved word, and you could have a `define` function in your code, for example. Said that, I believe that while traversing the AST, if we resort to scope detection when `define` is called we may probably infer that we are processing and AMD module if the `define` call is made in the global object context.

However, if we decide to do that, I think it would be better to leave this plugin as it is now, and create a new "automatic" one, so that people can choose one or the other. This is a more flexible approach as it is better to have simpler plugins that can be cherry picked, than having a huge intelligent plugin that does everything.

Also, this plugin would be much faster than the smarter one and, up to the date, we haven't find any scenario where mixed dependencies appear (at least for the mainstream frameworks we are testing).

yuchi:
The problem will arise when a misconfoguration of this build process in a marketplace app will break other apps.
In this scenario there's no place for configurability IMHHHHO (I'd like to stress on the humbleness of the opinion)

izaera:
Regarding that possibility, I think that an approach where shared packages are checked to verify and certify them would work much better. 

Even if you don't misconfigure a project you are building, there's always the possibility that you deploy a NPM package that is not the same as the one deployed in the NPM registry (for instance: you could have edited it by hand in your node_modules folder, or maybe downloaded it from a wrong place, whether on purpose or not). This would poison the entire portal in the same way as a badly configured packaging if your broken package is the one selected to be published by the portal.

To avoid that I thought about two possibilities (any other suggestions on ways to prevent it are welcome :-) ):

1. Create a way of prioritizing NPM packages that are deployed several times in different OSGi modules: that way anyone could deploy a bundle with the certified versions of the packages and set them to the maximum priority so that they cannot be overridden.

2. Check with NPM registry every deployed package to verify that they are not counterfeit (for example using a hash). This has the problem that the server would need to have access to Internet.

yuchi:
You can’t hash-compare anything since you (“you” as “liferay-npm-build-tools”) are tampering the contents in the first place.

yuchi:
Having a npm resources manager inside the portal could be an option but we have to guarantee

1. consistency and determinism between prod/staging environments and dev environments;
2. a way to export such configurations (as a lockfile of sort) shareable and installable as an OSGi bundle.

izaera:
Your 2 is almost trivial and is what I proposed in my 1. So, for example, it would be easy to create a tool that dumps an snapshot of a portal NPM packages to an OSGi bundle so that it can be delivered to other portal instances. 

Just brainstorming how it could look like...

izaera:
However we would still need to determine the validity of those NPM packages or provide a way to determine it. 🤔

PR #10, opened 2017-07-11
Remove suggestion for global installation of Lerna:

yuchi:
Since `lerna` is one of the `devDependencies` of the project there’s no need to have a globally installed instance too.

All command run through `npm run` will have `lerna` available anyway.


jbalsas:
You're right! Nothing escapes your inquisitive 👁 !

yuchi:
> I’m in the dark… searching for typos… you can’t escape my inquisitive 👁 !

PR #11, opened 2017-07-11
Add failing tests on packages that end with '.js'

yuchi:
When normalizing require names a package name must not be touched at all.

While discouraged, packages can still end with `.js` so removing it from the name would break the require call.

Also simply checking for a `/` in the require name is not enough because we need to include scoped packages in the equation, those in the form `@scope-name/package-name[/sub/module]`.

### DIFF
--- a/packages/babel-plugin-normalize-requires/src/__tests__/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin-normalize-requires/src/__tests__/__snapshots__/index.test.js.snap
@@ -15,6 +15,16 @@ exports[`when requiring external modules removes trailing "/" from module names 
 require('a-package/a-module.js');"
 `;
 
+exports[`when requiring local modules keeps trailing ".js" from package names 1`] = `
+"
+require('a-package.js');"
+`;
+
+exports[`when requiring local modules keeps trailing ".js" from package names 2`] = `
+"
+require('@some-scope/a-package.js');"
+`;
+
 exports[`when requiring local modules removes trailing ".js" from module names 1`] = `
 "
 require('./a-module');"
--- a/packages/babel-plugin-normalize-requires/src/__tests__/index.test.js
+++ b/packages/babel-plugin-normalize-requires/src/__tests__/index.test.js
@@ -2,6 +2,30 @@ import * as babel from 'babel-core';
 import plugin from '../index';
 
 describe('when requiring local modules', () => {
+	it('keeps trailing ".js" from package names', () => {
+		const source = `
+	    require('a-package.js')
+	    `;
+
+		const { code } = babel.transform(source, {
+			plugins: [plugin],
+		});
+
+		expect(code).toMatchSnapshot();
+	});
+
+	it('keeps trailing ".js" from package names', () => {
+		const source = `
+	    require('@some-scope/a-package.js')
+	    `;
+
+		const { code } = babel.transform(source, {
+			plugins: [plugin],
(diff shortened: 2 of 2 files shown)


yuchi:
Also please keep in mind that the current approach is too aggressive. If someone is requiring a folder which ends with `.js` it should not be stripped away.

Generally a better approach would be to resolve the extension at runtime, in the loader, having a list of exceptions to the rules. Please have a look at yuchi/frontend-ng-loader-workspace#5 on how i initially investigated the issue.

izaera:
@yuchi Good point: package names should be left untouched 👍 . The duality between packages and modules which makes them look the same (when they are not) makes me mad. And then, adding scopes to the equation makes it even funnier 😱.

Regarding `If someone is requiring a folder which ends with .js it should not be stripped away.`: that's what we are doing now if the user does a `require('folder.js/')` but it would fail if they did a `require('folder.js')`. I suppose that's what you mean, don't you?

BTW, thx for the feedback :-).

izaera:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

izaera:
I will close this PR, work on the feedback, and resend it again with all implemented.

izaera:
Thinking about the folder requires, I think that the normalizer should append `/index` to the module name, otherwise the AMD loader will fail. 

I think that normalizing requires (done right) is a safer approach than adding functionality to the AMD loader to handle aliases. The loader definitely handles aliases for the default entry point of packages because in that case the module name is configurable (by the `main` entry of the `package.json` file) but in the folder cases, appending `/index` at build time to requires should do the trick and avoid more complexity in both the loader and the NPM registry in the backend.

izaera:
OK, after reading the explanations in https://github.com/yuchi/frontend-ng-loader-workspace/issues/5, appending `/index` to folder requires does not solve the issue, we need to check for `package.json` files too. 

I'm going to create an issue for this to reflect more on it. I believe that we can normalize requires at build time with this plugin better than sending info to the loader to make it resolve module names at runtime, but I have to carefully think about it to make sure that we have all needed information always available.

yuchi:
I don't believe you'll be able to safely normalize require names at build time.

If a folder gets changed to a file in a minor release of a package we would have incorrectly normalized for a wrong file tree.

izaera:
That’s true. That’s what I was referring to when I said that I had to think if I had all the information *always* available. Your case proves I don’t ;-). 

As long as a package can be replaced by another one due to semver, its structure cannot be analyzed during build time. 

This will get more complex, then :-(.

yuchi:
The same applies for all changes that touch exceptions to the simple `+ '.js'` rule. That includes index (.js,.json) files in folders, package.json in folder, deep resolved main properties of package.json in folders, simple .json files required extensionless, files and folders that share the name

yuchi:
On the other hand the require name can be analyzed and resolved at serve time. But that would make this whole project partially useless.

BYW great stuff here. Could be interesting make it compatible with WebPack plugins. An experiment for another week.

izaera:
Of course, we could do all these operations in the portal, but we believe it would be much more complex to maintain (no Javascript support and slower release lifecycles) so we prefer to do as much as possible at build time :-).

Regarding webpack, we are releasing our own bundler because all standard bundlers target dependency resolution at build time and we delay it until page rendering so we are not sure that what we do today based on those standard bundlers won't break in the future.

PR #12, opened 2017-07-11
Don't remove trailing .js from package names

izaera:


izaera:
This is the continuation of https://github.com/liferay/liferay-npm-build-tools/pull/11

yuchi:
The PR name now doesn't reflect the changes (since it actually solves the issue!)

This is a better approach, but still not complete. See discussion on #11

jbalsas:
I'm merging for now since @yuchi said this is at least a better approach... 

Couldn't we normalize in the loader on require-time?

izaera:
@yuchi @jbalsas Please continue the discussions about this topic in this issue https://github.com/liferay/liferay-npm-build-tools/issues/14

thx :-)

PR #16, opened 2017-07-12
Add tests for liferay-npm-bundler

izaera:


jbalsas:
Hey @Robert-Frampton, this is the biggest thing we need to complete our framework/workflow story for Liferay Portal. Could you take a look at the architecture and provide some feedback?

@izaera, could you explain briefly here what this is supposed to do and why so Rob has some context?

izaera:
Explanation: this is a kind of bundler (the technical name for tools like browserify) that targets liferay-portal. Normal bundlers generate one large js file with all deps inside (some may split it in chunks, but conceptually it is just one file) but we don't because we resolve package dependencies in the very last moment: when rendering the page. This is because we need to mix several portlets which, in the end, are like tiny nearly-isolated web apps. 

Said that, when we package an OSGi bundle JAR, we:

1. Put all NPM dependencies (only the ones really needed by the project) inside the JAR
2. Pre process each dependency package with npm-bundler plugins
3. Process each dependency package by running Babel on all source files and copying the rest verbatim
4. Post process each dependency package with npm-bundler plugins

And this is the tool that does those things.

The rationale behind 2, 3, and 4 is that 2 prepares the package before running Babel, 3 runs Babel to wrap modules inside AMD defines, normalize things, shim things, and the like, and 4 polishes Babel's output if necessary.

Hope this explains everything and makes it clear like a crystall ball ;-).

Robert-Frampton:
Hey @izaera 

What's the best way to examine what the output looks like? Is there a test module you've been using?

izaera:
@Robert-Frampton Sorry for not answering. Looks like I missed the notification :-(.

There's this repo with examples: https://github.com/izaera/liferay-npm-examples . You can build the projects in there and examine the generated JARs. There are examples for several libraries/frameworks (react, vue, angular, ...).

Robert-Frampton:
Hey @izaera 

Awesome, I'll take a look today.

Thanks!

izaera:
@Robert-Frampton Did you manage to find some time to look at this?

PR #23, opened 2017-08-21
Add documentation for babel plugins

izaera:


packages/babel-plugin-name-amd-modules/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+For example, given a `srcPrefixes` of 
+`["src/main/resources/META-INF/resources"]` if the module under 
+`src/main/resources/META-INF/resources/lib/index.js` is processed, its relative 
+path is computed to `lib/index.js`.
+
+The default value of the `srcPrefixes` option is 
+`["src/main/resources/META-INF/resources"]`.
+
+## Example
```
jbalsas: Can you verify this is being rendered as a title properly? Didn't look like it on the GitHub preview...

packages/babel-plugin-name-amd-modules/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+**In**
+
+```
```
jbalsas: All code blocks are missing the language to enable syntax highlighting. Can you review them adding `javascript`, `sh` or `json` where appropriate?

packages/babel-plugin-name-amd-modules/README.md
```diff
@@ -0,0 +1,74 @@
+# babel-plugin-name-amd-modules
+
+> Give a name to AMD modules based on package name and version and module path.
+
+## Details
```
jbalsas: Some of the `details` section become too verbose pushing down the more easy to understand examples...

Maybe we could try https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d so at least it shows collapsed on a glimpse through the GitHub interface? Other option would be to push details down, but I'm not sure how much sense does that make.

What do you think?

izaera: I would prefer moving them down and maybe changing the title from Details to Technical Details or something like that. Using <details> sounds good, but it only works with github and that kind of abuses the MD format (I feel).

jbalsas: The `Technical Details` idea sounds good to me 👍

PR #28, opened 2017-08-23
Add documentation for bundler presets

izaera:


izaera:
This was already sent in #26

PR #33, opened 2017-08-25
liferay-npm-bundler fails with a missing module error.

izaera:


jbalsas:
Hey @izaera, I've merged so we can push forward, but I still think we should try to simply use the `mkdirp` npm module instead of our own implementation if possible. We can revisit later after the release 😉

PR #62, opened 2017-11-14
Load Babel configuration when specified by package name (without id)

izaera:


yuchi:
Should a semver range be a better approach at targeting packages?

```
*         # every package, at every version
xx        # every version of 'xx'
xx@*      # every version of 'xx'
xx@^1.2.3 # every version of 'xx' that matches '^1.2.3'
xx@3.4.5  # 'xx' at version '3.4.5'
```

You could simplify the rules a lot, giving even more power with simple additions. You cpuld split by `@` (keep scoped pkgs in mind) and extract package name and version range, defaulting **both** to `*`. Then treat the `name` as a glob (`babel-*`) and the `version` as semver range.

izaera:
@yuchi Sounds good, though we haven't needed it yet. I'll add it as an enhancement issue for the future. Thx 👍 .

PR #89, opened 2018-03-15
add null check

gamerson:


gamerson:
hey @izaera  i would be willing to add an integration test for this, but I'm not sure how! :)

izaera:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

izaera:
> hey @izaera i would be willing to add an integration test for this, but I'm not sure how! :)

No worries. We don't have integration tests for the bundler yet as they would be very tricky to set up and we are not sure they are worth the effort (in this moment). However, we have unit tests for the rest of the setup, and believe that's sufficient for now.

Anyway, thanks for the contribution :-)

PR #93, opened 2018-04-11
Fix CI tests

izaera:


izaera:
Will resend with better comments

PR #105, opened 2018-04-24
Add a config option to ignore .js files when the bundler processes them

izaera:


packages/liferay-npm-bundler/src/runners.js
```diff
@@ -72,12 +73,21 @@ export function runBabel(pkg) {
...
+		// Filter nested node_modules files
 		const nodeModulesPrefix = path.resolve(`${pkg.dir}/node_modules`);
-
 		filePaths = filePaths.filter(
 			filePath => !filePath.startsWith(nodeModulesPrefix)
 		);
 
+		// Filter ignore files
+		const outputPrefix = path.resolve(pkg.dir);
+		filePaths = filePaths.filter(
```
jbalsas: Isn't there a way to to this through the `globby` API? so we only get the proper filepaths in here already, for example, something like:

```javascript
return globby([
    `${pkg.dir}/**/*.js`,
    `!${ignore}`
]).then(...
```

izaera: Maybe. Let me check.

jbalsas: Seeing `ignore` is an array...

```
const globs = [`${pkg.dir}/**/*.js`].concat(
    ignore.map(glob => `!${glob}`
);
```

PR #107, opened 2018-04-26
Add tests for inject-imports plugin

izaera:


packages/liferay-npm-bundler/src/index.js
```diff
@@ -1,5 +1,6 @@
 import fs from 'fs-extra';
 import globby from 'globby';
+import * as gl from 'liferay-npm-build-tools-common/lib/globs';
```
jbalsas: Why import * if we just need `negate`?

jbalsas: I stand corrected, we use both :D

PR #166, opened 2018-07-10
Add create-jar option to liferay-npm-bundler

izaera:


packages/liferay-npm-bundler/src/config.js
```diff
@@ -232,6 +239,14 @@ export function getBabelConfig(pkg) {
 	return getPackageConfig(pkg, '.babelrc', {});
 }
 
+/**
+ * Whether or not to create an OSGi bundle
+ * @return {boolean}
+ */
+export function isCreateJar() {
```
jbalsas: Since this is to `whether or not to create an OSGi bundle`... would it make more sense to call this `isCreateOSGiBundle` or something like that?

A `jar` does not necessarily need an OSGi bundle, so it looks like we're going to do a bit more than just a jar

jbalsas: Ignored by @izaera after an offline conversation about this :)

PR #182, opened 2018-09-07
Fixes #179 - Add auto-deploy-portlet option support

### LINKED ISSUE #179: Create an option in .npmbundlerrc to disable extender when create-jar is true
This is necessary for shared module projects (for example: an OSGi bundler exporting Angular) where no portlet should be created.

The option should just stop the bundler from adding the `Require-Capability: osgi.extender;filter:="(osgi.extender=liferay.frontend.js.portlet)"` line to MANIFEST.MF.

izaera:


izaera:
Wrong branch :-(

PR #185, opened 2018-09-07
Fixes #174 - Create generator for shared packages bundle

### LINKED ISSUE #174: Create Yeoman Generator for NPM Portlet Extender
The new NPM portlet extender simplifies Portlet development for Frontend developers.

Creating a generator to bootstrap the most common use cases would greatly lower the entry barrier to create a widget.

The generator should:
- Allow to choose between the main "supported" frameworks
    - React
    - Vue.js
    - Angular
    - Metal.js
    - Standalone
- Walk through the possible portlet attributes that can be added
- Include a scaffolding standalone html page so a developer could simply watch the app being made without a Portal instance

izaera:


packages/generator-npm-portlet-extender/src/config.js
```diff
@@ -0,0 +1,16 @@
+import os from 'os';
+import path from 'path';
+import readJsonSync from 'read-json-sync';
+
+const cfg = readJsonSync(path.join(os.homedir(), '.npm-portlet-extender.json'));
```
jbalsas: This will stop execution if the file is missing

jbalsas:
Hey @izaera, `liferay-bundle` sounds like the most promising... 🤔 

Regarding the html and css, what I mean is keeping all the bootstrapping and wiring, but just have the view render nothing, no html printing the contextPath or anything. In any case, this is a minor thing that might help someone get started...

jbalsas:
Looks good! Merging so we can keep going on top of this!

I think we should wait until we merge in the local-dev story before publishing an initial canary version.

PR #236, opened 2018-11-19
Fixes #231 - Modify generator to stop using legacy configuration

### LINKED ISSUE #231: Refactor configuration management
With the advent of JAR creation, configuration file and management is becoming a huge mess.

We should refactor the code and reorganize the .npmbundlerrc so that it is more organized in sections. At least we should divide the config in these subparts:

* base global config
* babel config
* bundler config
* jar creation config

izaera:


jbalsas:
LGTM! I'd just add auto-detect for the localization and similar features, so writing it explicitly in the conf is not necessary.

Additionally, for some time, we could revisit the access pattern to config properties, so we can avoid making all the checks and resolve chained properties like `get(jarConfig, 'features.location.default.values')`.

packages/liferay-npm-bundler/src/config/_util.js
```diff
@@ -0,0 +1,57 @@
...
+ * name.
+ * @param {PkgDesc} pkg the package descriptor
+ * @param  {String} section the section name (like 'plugins', '.babelrc', ...)
+ * @param  {Object} defaultValue default value if not configured
+ * @return {Object} a configuration object
+ */
+export function getPackageConfig(pkg, section, defaultValue = undefined) {
+	let pkgConfig;
+
+	if (config.packages[pkg.id] && config.packages[pkg.id][section]) {
```
jbalsas: I saw you introduced [dot.prop](https://www.npmjs.com/package/dot-prop). Why not use it here?

izaera:
> Hey @izaera, what's the meaning of the _ in front of file names? What are you trying to signal with that?

I mean those packages are not intended to be consumed from outside the `config` facade (private). We can use any other convention if we don't like that (like putting them in `internal` or something like that...).

packages/liferay-npm-bundler/src/config/_util.js
jbalsas: got it

jbalsas:
Let's move it to an `internal` folder, like https://github.com/nodejs/node/tree/master/lib/internal 🤷‍♂️

PR #240, opened 2018-11-23
Fixes #233 - SF

### LINKED ISSUE #233: Modify yeoman generators to support localization
Adapt yeoman generator to changes in the following issues:

* #231  
* #232

izaera:


packages/generator-liferay-bundle/src/target-react-portlet/templates/src/AppComponent.js.ejs
```diff
@@ -6,15 +6,15 @@ export default class extends React.Component {
 		return (
             <div>
 				<div>
-					<span className="tag">Portlet Namespace:</span> 
+        	        <span class="tag"><%- labels.porletNamespace %>:</span>
```
jbalsas: Wouldn't here work simply as `Liferay.Language.get('portletNamespace')`?

packages/generator-liferay-bundle/src/target-angular-portlet/angular.js
```diff
@@ -41,28 +42,55 @@ export default class extends Generator {
...
-
+		// Configure webpack
 		pkgJson.addDevDependency('ts-loader', '^5.0.0');
 		npmbuildrc.addWebpackRule(/src\/.*\.ts$/, 'ts-loader');
 		npmbuildrc.addWebpackExtensions('.ts', '.js');
 		npmbuildrc.setWebpackMainModule('index.ts');
 
+		// Prepare text labels
+		const labels = formatLabels({
+			portletNamespace: 'Porlet Namespace',
```
jbalsas: Keys are usually `portlet-namespace`... at least in Liferay. Is that a Liferay-only convention?

packages/generator-liferay-bundle/src/target-react-portlet/templates/src/AppComponent.js.ejs
izaera: L10N is optional so we need to support two scenarios: one calling `Liferay.Language.get`, and another writing the raw string so that the developer can then decide how to localize the app. Maybe using his own framework tools...

packages/generator-liferay-bundle/src/target-angular-portlet/angular.js
izaera: That's our convention yes, but you can use whatever you like as keys. We even use a mix of dots and hyphens sometimes.

izaera:
> One question though... are we actually generating a real Language.properties file? Those here seem empty, but I guess the generator fills them up?

Yes, the generator fills up the Language.properties files as needed, depending on the answers of the developer running the generator.

packages/generator-liferay-bundle/src/target-angular-portlet/angular.js
jbalsas: I would at least follow our convention for the generated keys... just so it doesn't look weird to those used to it. If they want to then change them, they'll be free to do so

PR #243, opened 2018-12-05
Fixes #242 - Let SDK version be specified in the home config file while still using QA configs

### LINKED ISSUE #242: Refactor QA tests
With the increasing casuistry, the QA tests need more automation.

izaera:


packages/generator-liferay-bundle/src/facet-settings/index.js
```diff
@@ -0,0 +1,59 @@
...
+		const npmbundlerrc = new NpmbundlerrcModifier(this);
+		const projectAnalyzer = new ProjectAnalyzer(this);
+
+		npmbundlerrc.setFeature('settings', DEFAULT_SETTINGS);
+
+		let context = {
+			name: '',
+		};
+
+		if (!projectAnalyzer.hasLocalization && projectAnalyzer.description) {
```
jbalsas: Why does the settings facet need to know about localization?

packages/generator-liferay-bundle/src/facet-settings/sample-generator.js
```diff
@@ -0,0 +1,63 @@
...
+	/**
+	 *
+	 */
+	generate() {
+		const projectAnalyzer = new ProjectAnalyzer(this._generator);
+
+		if (projectAnalyzer.hasSettings) {
+			const settingsJson = new SettingsJsonModifier(this._generator);
+
+			if (projectAnalyzer.hasLocalization) {
```
jbalsas: Same doubt that above... this makes me think we probably need some type of AOP or facet lifecycle down the line, so for example the localization facet could take care of this.

packages/generator-liferay-bundle/src/utils/modifier/features/settings.json.js
```diff
@@ -0,0 +1,48 @@
+import prop from 'dot-prop';
+
+import {JsonModifier} from '../..';
```
jbalsas: I assume `JsonModifier` is the `index.js` one here? Maybe we could name it, or do we want it to be the default entry point for some other reason?

packages/liferay-npm-bundler/src/jar/__tests__/xml.test.js
```diff
@@ -1,27 +1,92 @@
...
+	it('works with just the type', () => {
+		const xml = createMetatype('id', 'name');
+		addMetatypeAttr(xml, 'an-attr', {
+			type: 'string',
+		});
 
-describe('patchMetatypeXml', () => {
-	it('works with localization file', () => {
-		const xmlFile = path.join(__dirname, '__fixtures__', 'metatype.xml');
+		expect(xml).toMatchSnapshot();
```
jbalsas: Some of these tests here are a bit confusing now, since we're saying `xml`,  but we're actually generating `json`, aren't we?

packages/liferay-npm-bundler/src/jar/__tests__/xml.test.js
```diff
@@ -1,27 +1,92 @@
...
+	addMetatypeAttr(xml, 'an-option', {
+		type: 'string',
+		default: 'A',
+		options: {
+			A: 'option-a',
+			B: 'option-b',
+		},
+	});
+
+	expect(format(xml)).toMatchSnapshot();
```
jbalsas: This would be the only one to convert the settings to xml 👍

packages/generator-liferay-bundle/qa/generate-samples.js
```diff
@@ -6,11 +8,88 @@ const rimraf = require('rimraf');
...
+		JSON.stringify(
+			{
+				batchMode: true,
+				answers: {
+					'*': Object.assign(
+						{
+							description: options.folder,
+							category: 'category.sample',
+							liferayPresent: true,
+							liferayDir: '/Users/ivan/Liferay/CE/bundles',
```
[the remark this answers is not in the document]
izaera: Oh yes, that's true. I saw it when testing but forgot to remove it :-S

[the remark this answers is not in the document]
izaera: Pushed a modification

packages/generator-liferay-bundle/qa/generate-samples.js
```diff
@@ -6,11 +8,88 @@ const rimraf = require('rimraf');
...
 fs.mkdirSync(outDir);
 fs.mkdirSync(pkgsDir);
 
+const start = new Date();
+
+// Generate vanilla samples configuration
+[true, false].forEach(useBabel => {
+	[true, false].forEach(useLocalization => {
+		[true, false].forEach(useSettings => {
+			[true, false].forEach(sampleWanted => {
```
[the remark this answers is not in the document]
izaera: Let's see how this evolves. It's very probable that if we have more options we won't be able to generate so many samples, so we won't need to enhance this code... 🤷‍♂️

packages/generator-liferay-bundle/src/config.js
```diff
@@ -5,15 +5,20 @@ import readJsonSync from 'read-json-sync';
...
 
 try {
-	let configPath = path.join(os.homedir(), '.generator-liferay-bundle.json');
+	cfg = readJsonSync(
+		path.join(os.homedir(), '.generator-liferay-bundle.json')
+	);
 
 	const argv = process.argv;
 
 	if (argv.length >= 5 && argv[3] === '--config') {
```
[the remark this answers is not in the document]
izaera: But yargs is for arguments 🤔. 

The idea of having a `.generator-liferay-bundle.json` in your $HOME is that you write it once with predefined defaults (for example: where's liferay dir, never write samples, use always react, and things like that...) and forget about it. 

It should represent your development environment and assumptions, that's why it is located at $HOME.

If we need multiple development environments or assumptions in the future we can change it, but I wouldn't bother too much with these because it can be also managed from outside of the tool, by carefully generating/defining config files. 

Also, it is very unlikely that users of the tool want to use this file because people won't be usually generating lots of projects lots of days. This is more for internal comsumption when running the QA tests than for real life scenarios unless someone finds a use case where it is very useful for him which, as of now, hasn't happened.

packages/generator-liferay-bundle/src/facet-settings/index.js
izaera: Because settings labels can be localized or not, so we need to check if localization is active when generating the config. Of course, we could make settings facet less smart and always output unlocalized strings, but this way it's more powerful and less error prone, I hope.

packages/generator-liferay-bundle/src/facet-settings/sample-generator.js
izaera: The way it is organized now is hierarchical. Localization "runs" before settings, sample generation, and everything else. That means that it is the responsibility of settings and sample facets (note that there's no sample facet per se, it's embedded inside targets) to test if L10N is active to act accordingly.

This of course implies that if you have a project with settings or samples which was created unlocalized and you then add localization, the settings and the samples won't be localized by the facet and you will need to do it by hand.

We could find a way to do some cross referenced algorithms so that it doesn't matter the order in which you apply facets, but I doubt it would work because, once you have generated your project and modified it, it's very difficult for any generator to know how to modify certain things. It's maybe easy for config files, but for code it will be another magnitude of complexity, I think.

But it's something definitely worth investigating. We haven't even started to run facets after project generation to see how it works and what can be achieved, so it's very premature to think about this, I think.

packages/generator-liferay-bundle/src/utils/modifier/features/settings.json.js
izaera: That's a result of how the code evolved. There are more things inside `index.js`. But yes, I'll move it to its own file. It's definitely ugly like it is now.

packages/liferay-npm-bundler/src/jar/__tests__/xml.test.js
izaera: Mmm, not really. Even though they are JS objects (not JSON) they are in-memory representations of XML files. It's like DOM model, but simpler. That's why I refer to those objects as XML.

It's the `format()` function who does the magic of converting the in-memory XML to an actual XML string.

We could rename it to `serialize()` too, if it seems clearer...

packages/liferay-npm-bundler/src/jar/__tests__/xml.test.js
izaera: To an XML string yes. But we need to differentiate between the input JSON file (the one the developer writes) and the in-memory XML file (the output metatype XML).

packages/generator-liferay-bundle/src/config.js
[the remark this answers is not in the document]
jbalsas: But why is this positional? `argv[3] == '--config'`? What does that mean?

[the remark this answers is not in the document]
izaera: Oops, I misunderstood your comment. Thought it was referring to the changes instead of the last line :-). OK, I'll create an issue to investigate the possibility of using yargs or making that code more robust 👍 .

PR #252, opened 2019-01-17
Correct spelling of JavaScript

wincent:
Noticed this during @izaera's walkthrough this morning.


wincent:
"master" and "develop" currently point at the same thing, but per @izaera's in-person comment I just changed the base to the "develop" branch so it will land there first.

PR #263, opened 2019-02-15
Add support for portlet preferences in JS-extended portlets

izaera:


packages/generator-liferay-bundle/qa/setup.sh
```diff
@@ -20,6 +20,13 @@ if [ $? != 0 ] ; then
     exit 1
 fi
 
+which lerna > /dev/null
```
wincent: Should probably use `command -v` for this: https://stackoverflow.com/a/677212/2103996

packages/generator-liferay-bundle/src/target-angular-portlet/index.js
```diff
@@ -1,28 +1,24 @@
 import Generator from 'yeoman-generator';
 
+import * as standardTarget from '../utils/target/standard';
+
 /**
  * Generator for Angular portlets.
  */
-export default class extends Generator {
+export default class ThisGenerator extends Generator {
```
wincent: What's the meaning behind the naming of `ThisGenerator`?

packages/generator-liferay-bundle/src/utils/ProjectAnalyzer.js
```diff
@@ -117,6 +131,29 @@ export default class ProjectAnalyzer {
...
 	}
 
+	/**
+	 * Get the path to the preferences file.
+	 * @return {string}
+	 */
+	get preferencesFilePath() {
+		const fs = this._generator.fs;
+
+		let preferences = prop.get(
```
wincent: Seems this `let` should be `const`.
```suggestion
		const preferences = prop.get(
```

packages/generator-liferay-bundle/src/utils/ProjectAnalyzer.js
```diff
@@ -117,6 +131,29 @@ export default class ProjectAnalyzer {
...
+			this._npmbundlerrc,
+			'create-jar.features.preferences'
+		);
+
+		if (preferences) {
+			return preferences;
+		} else {
+			if (fs.exists(DEFAULT_PREFERENCES)) {
```
wincent: `else if` here would allow you to reduce the nesting, which is usually a good thing.

packages/generator-liferay-bundle/qa/setup.sh
izaera: No need to be so precise in these scripts as they are only for very internal use, but interesting link, nonetheless :-).

packages/generator-liferay-bundle/src/target-angular-portlet/index.js
izaera: It's a leftover of a spike :-S. Will remove it.

packages/generator-liferay-bundle/src/utils/ProjectAnalyzer.js
izaera: :+1:

izaera:
Doesn't show the new commits, so I'll resend

PR #272, opened 2019-02-22
Add support for configuration categorization

izaera:


packages/liferay-npm-bundler/src/jar/index.js
```diff
@@ -188,6 +189,30 @@ function addMetatypeFile(zip) {
...
+
+	const metatypeJson = {};
+
+	if (json.category) {
+		metatypeJson.category = json.category;
+	}
+
+	zip
+		.folder('features')
+		.file(`metatype.json`, JSON.stringify(metatypeJson, null, 2));
```
wincent: Style nit: you used single quotes on the previous line and backticks (without interpolation) on this line, but should probably use the same quoting style for both.

izaera: :+1:

PR #273, opened 2019-02-25
Only print the warning if the verbose option is true

julien:


izaera:
@julien  The commit looks OK, but it is missing the "Fixes #xxx" description.

Can you describe the issue that this commit refers to? If there's no issue, create one, assign it the "enhancement" label and the milestone (I suppose it's OK to set it to "next-version", but if it is very urgent we can merge it in master and release a maintenance version).

Once that's done, refer to the issue in the commit message.

Thx :+1:

julien:
@izaera thanks, I re-sent it here #275

PR #277, opened 2019-02-28
Fixes #262 | Modify projects generation to account for the new configuration format

### LINKED ISSUE #262: Reshape 'configuration' parameter of entry point
> 👀 THIS WILL INTRODUCE A BREAKING CHANGE

However, as the feature has been unusable until now, because the JS Portlet Extender was not released, this breaking change shouldn't be a big problem.

Nevertheless, we will try to remove 2.6.x versions from npmjs so that nobody gets confused in the future in case their version numbers are messed.

----
This issue depends on [LPS-91364](https://issues.liferay.com/browse/LPS-91364)

The goal of this issue is to reshape the `configuration` parameter to make it look like:

```javascript
configuration: {
	system: {...},
	portletInstance: {...}
}
```

Where `system` will hold the configuration defined in `Control panel` > `System settings` and `portletInstance` will hold the portlet preferences as defined in the portlet spec.

To maintain coherence, we will also merge the files that describe these configuration schemas into just one single file named `configuration.json` with the following structure:

```json
{
	"system": {
		"fields": {
			"fruit": {
				"type": "string",
				"name": "fruit",
				"description": "fruit-help",
				"default": "orange",
				"options": {
					"orange": "an-orange",
					"pear": "a-pear",
					"apple": "an-apple"
				}
			}
		}
	},
	"portletInstance": {
		"fields": {
			"fruit": {
				"type": "string",
				"name": "fruit",
				"description": "fruit-help",
				"default": "orange",
				"options": {
					"orange": "an-orange",
					"pear": "a-pear",
					"apple": "an-apple"
				}
			}
		}
	}
}
```

So, where you previously had a `settings.json` file, you must move its contents to a single `configuration.json` file and wrap them inside the `system` section. Then update `.npmbunderrc` accordingly.

----

This will supersede the following issues:

* #234 
* #241 
* #260 
* #270 

----

It will also render the following versions unusable:

* [2.6.0](https://github.com/liferay/liferay-npm-build-tools/issues?q=is%3Aclosed+milestone%3A2.6.0)
* [2.6.1](https://github.com/liferay/liferay-npm-build-tools/issues?q=is%3Aclosed+milestone%3A2.6.1)
* [2.6.2](https://github.com/liferay/liferay-npm-build-tools/issues?q=is%3Aclosed+milestone%3A2.6.2)

All people using them will need to switch to [2.7.0](https://github.com/liferay/liferay-npm-build-tools/issues?q=is%3Aclosed+milestone%3A2.7.0).

Also all projects containing configuration definitions will need to update their `.npmbundlerrc` files to use the new structure.

izaera:


packages/generator-liferay-bundle/src/facet-configuration/sample-generator.js
```diff
@@ -0,0 +1,116 @@
+import ProjectAnalyzer from '../utils/ProjectAnalyzer';
+import ConfigurationJsonModifier from '../utils/modifier/features/configuration.json';
+import LanguagePropertiesModifier from '../utils/modifier/features/localization/Language.properties';
+
+/**
+ *
```
wincent: Can we supply comment content here or delete the empty comment? (Same for other empty comment below)

packages/generator-liferay-bundle/src/facet-configuration/sample-generator.js
```diff
@@ -0,0 +1,116 @@
...
+							water: 'water',
+							wine: 'wine',
+							beer: 'beer',
+						},
+					}
+				);
+
+				// Add configuration localization
+				new LanguagePropertiesModifier(this._generator).addProperties({
+					'fruit': 'Favourite fruit',
```
wincent: en_US spelling would be "Favorite"

packages/generator-liferay-bundle/src/facet-configuration/sample-generator.js
```diff
@@ -0,0 +1,116 @@
...
+					'string',
+					{
+						name: 'drink',
+						description: 'drink-help',
+						required: false,
+						defaultValue: 'water',
+						options: {
+							water: 'water',
+							wine: 'wine',
+							beer: 'beer',
```
wincent: cokezero

packages/generator-liferay-bundle/src/utils/ProjectAnalyzer.js
```diff
@@ -109,49 +109,28 @@ export default class ProjectAnalyzer {
...
 		);
 
-		if (settings) {
-			return settings;
+		if (configuration) {
+			return configuration;
 		} else {
-			if (fs.exists(DEFAULT_SETTINGS)) {
-				return DEFAULT_SETTINGS;
+			if (fs.exists(DEFAULT_CONFIGURATION)) {
```
wincent: Could reduce nesting here with `else if`

packages/generator-liferay-bundle/src/utils/target/standard.js
```diff
@@ -25,16 +23,19 @@ export function initializing(generator) {
  * @param {string} namespace
  */
 export async function prompting(generator) {
-	generator.answers == generator.answers || {};
```
wincent: 😱

izaera: I'm sure I changed this. :thinking:

izaera: Don't know why it appears in the old way...

wincent: Yeah you did. I was just 😱 about the fact that it was in there in the first place. Good example of a thing that I would want ESLint to complain about (expression without a side-effect) instead of some of the other stuff it complains about that is not useful...

izaera: Ah OK. Let me look if I can add a rule for that...

wincent: > Don't know why it appears in the old way...

It's just GitHub UI fail. If you look in side-by-side mode then it's obvious that the comment applies to the old version and not the new one:

banners_and_alerts_and_fixes__262___modify_projects_generation_to_account_for_the_new_configuration_format_by_izaera_ _pull_request__277_ _liferay_liferay-npm-build-tools

packages/generator-liferay-bundle/src/facet-configuration/sample-generator.js
izaera: We leave this for the future as we have to decide whether to disable the eslint rule.

packages/generator-liferay-bundle/src/utils/target/standard.js
izaera: Added :+1:

PR #287, opened 2019-03-18
Fixes #286 | Point generated configuration.json files to schema

### LINKED ISSUE #286: Create a JSON schema file for configuration.json
Creating a JSON schema for `configuration.json` would make it easier to integrate with autocompletion in IDEs and validation of `configuration.json` files.

izaera:


packages/generator-liferay-bundle/src/facet-configuration/templates/features/configuration.json.ejs
```diff
@@ -1,4 +1,5 @@
 { 
+	"$schema": "https://raw.githubusercontent.com/liferay/liferay-npm-build-tools/master/resources/schemas/configuration.schema.json",
```
jbalsas: I don't think we can rely on `raw.githubusercontent` for this. Is that standard practice?

Since we're publishing to npm, I think we might be able to leverage some of the existing cdns like [unpkg](https://unpkg.com/) or [jsDelivr](https://www.jsdelivr.com/)?

izaera: I've removed the las commit referring to this so that we can point the generator file to schemastore later.

PR #295, opened 2019-03-26
Improve merge strategy with presets

julien:


packages/liferay-npm-bundler/src/config/index.js
```diff
@@ -333,7 +334,10 @@ function loadConfig() {
 
 	if (presetFile) {
 		const originalConfig = Object.assign({}, config);
-		Object.assign(config, readJsonSync(presetFile), originalConfig);
+		Object.assign(
+			config,
+			merge.recursive(true, readJsonSync(presetFile), originalConfig)
+		);
```
wincent: Presumably, `readJsonSync` returns a new object every time, so there is no need to pass the `true` (ie. "clone") parameter to `merge.recursive()`, is there?

Wondering, would the following be clearer, seeing as `merge.recursive()` basically *is* `Object.assign()`, but on steroids?

```javascript
const originalConfig = Object.assign({}, config);
merge.recursive(
  false,
  config,
  readJsonSync(presetFile),
  originalConfig
);
```

packages/liferay-npm-bundler/src/config/__tests__/index.test.js
```diff
@@ -359,3 +359,72 @@ describe('jar config', () => {
 		});
 	});
 });
+
+describe('presets', () => {
+	it('should work with existing presets', () => {
+		process.chdir(
+			path.join(__dirname, '__fixtures__', 'config', 'presets')
+		);
```
wincent: `process.chdir()` probably belongs in a `beforeEach()` callback and be paired with a balancing `afterEach()` that restores the cwd. That way, if a test fails (or if it passes!), we always wind up back where we started.

packages/liferay-npm-bundler/src/config/index.js
julien: It certainly seems clearer ... but when passing `false` to `merge.recursive` the original values from the `originalConfig` object are not over written.

So I'd probably go with 

```
const originalConfig = Object.assign({}, config);
merge.recursive(
  true,
  config,
  readJsonSync(presetFile),
  originalConfig
);
```

packages/liferay-npm-bundler/src/config/__tests__/index.test.js
julien: Yes it does. Thanks

packages/liferay-npm-bundler/src/config/index.js
wincent: Interesting. I haven't used `merge` before, but I thought `true` just meant "start with a clone of the first object".  I guess this is because even though you make a "copy" in `originalConfig`, it is only a shallow copy (because it is made with `Object.assign()`). So, as the merge proceeds, the values that are deeply written into `config` end up changing the nested contents (below the top level) of `originalConfig`.

packages/liferay-npm-bundler/src/config/__tests__/index.test.js
julien: @wincent I should also probably do that for all the other tests that have this "pattern", but I think it might be wiser to do that in another PR.

wincent: Good call. I didn't even see the other tests (I'm only looking at this on GitHub, which doesn't give me a full view unless I click through.)

packages/liferay-npm-bundler/src/config/index.js
julien: Actually the snippet I added in my last comment does not work as expected 😞 

And you got that right.

julien: I removed the first argument to the `merge.recursive` call though, as it was not needed

wincent: Those optional-first-arg APIs are always a bit surprising. https://github.com/yeikos/js.merge/blob/b31e67fe6592390c967c991aa604c06ed2ae8c4f/merge.js#L25-L36

packages/liferay-npm-bundler/package.json
```diff
@@ -24,6 +24,8 @@
 		"jszip": "^3.1.5",
 		"liferay-npm-build-tools-common": "2.6.2",
 		"liferay-npm-bundler-preset-standard": "2.6.2",
+		"liferay-npm-bundler-preset-liferay-dev": "1.1.3",
```
wincent: Should this be in "devDependencies" as it is only referenced in the tests? (And then, only in a fixture.)

julien: Exactly. Thanks

PR #299, opened 2019-04-01
Fixes #298 | Bundler is failing with EMFILE errors

### LINKED ISSUE #298: Bundler is failing with EMFILE errors
Bundler is sometimes failing with EMFILE errors for packages with lot of files. This happens even if `process-serially` is set to true (see #50).

izaera:


packages/liferay-npm-bundler/src/config/internal/bundler.js
```diff
@@ -80,11 +80,19 @@ export function getPlugins(phase, pkg) {
...
- * @return {boolean}
+ * Get maximum number of files to process in parallel.
+ * @return {number}
  */
-export function isProcessSerially() {
-	return prop.get(config, 'process-serially', false);
+export function getMaxParallelFiles() {
+	// Default values for "ulimit -n" vary across different OSes. Some values
+	// I have found are:
+	//   - Apparently Mac OS X limit is 256 but usually people increases it
```
wincent:
```suggestion
	//   - Apparently Mac OS X limit is 256 but usually people increase it
```

packages/liferay-npm-bundler/src/config/internal/bundler.js
```diff
@@ -80,11 +80,19 @@ export function getPlugins(phase, pkg) {
...
+	//   - Fedora: 1024
+	//   - Windows: there's no ulimit, but MSVCRT.DLL has a 2048 limit
+	// Given this mess and the impossibility to retrieve the limit from Node,
```
wincent:
```suggestion
	// Given this mess and the impossibility of retrieving the limit from Node,
```

packages/liferay-npm-bundler/src/runners.js
```diff
@@ -85,102 +85,131 @@ export function runBabel(pkg, {ignore = []} = {}) {
...
-								pkg,
-								packageFilePath,
-								logger
+	const filePaths = globby.sync(globs);
+
+	return processBabelFiles(filePaths, 0, pkg, babelConfig);
+}
+
+/**
+ * Recursively process Javascript files with Babel chunk by chunk, to maintain
```
wincent:
```suggestion
 * Recursively process JavaScript files with Babel chunk by chunk, to maintain
```

packages/liferay-npm-bundler/src/runners.js
```diff
@@ -85,102 +85,131 @@ export function runBabel(pkg, {ignore = []} = {}) {
...
+ * an upper bound in the maximum number of open files so as to avoid EMFILE
```
wincent:
```suggestion
 * an upper bound on the maximum number of open files so as to avoid EMFILE
```

packages/liferay-npm-bundler/src/runners.js
```diff
@@ -85,102 +85,131 @@ export function runBabel(pkg, {ignore = []} = {}) {
...
+	);
+
+	const promises = filePathsChunk.map(
+		filePath =>
+			new Promise(resolve => {
+				const logger = new PluginLogger();
+
+				babelIpc.set(filePath, {
+					log: logger,
+					rootPkgJson: readJsonSync('package.json'),
```
wincent: Can we hoist this out of the loop?

You could also hoist the `config.getGlobalConfig()` below, right?

packages/liferay-npm-bundler/src/runners.js
```diff
@@ -85,102 +85,131 @@ export function runBabel(pkg, {ignore = []} = {}) {
...
+					globalConfig: config.getGlobalConfig(),
+				});
+
+				const packageFilePath = pkg.isRoot
+					? filePath.substring(
+							path.resolve(config.getOutputDir()).length + 1
+					  )
+					: filePath.substring(
+							filePath.indexOf(`${pkg.id}`) + pkg.id.length + 1
```
wincent:
```suggestion
							filePath.indexOf(pkg.id) + pkg.id.length + 1
```

packages/liferay-npm-bundler/src/runners.js
izaera: In fact no :sweat_smile: : what we want to do is to clone the `globalConfig`, in addition to the package.json object to isolate plugins from the core. Otherwise, plugins may modify our config or package.json and it would be passed corrupt to the other plugins.

I'm not talking about a remote possibility (at least for package.json) because it's very easy that a plugin modifies those objects.

PR #301, opened 2019-04-03
Fix #293: Add translate target to generated projects

### LINKED ISSUE #293: Add support for Microsoft Translator in generated projects
See https://issues.liferay.com/browse/IFI-570 also

izaera:


packages/liferay-npm-build-support/src/scripts/translate.js
```diff
@@ -0,0 +1,297 @@
...
+		process.exit(2);
+	}
+
+	const localizationFiles = cfg.getLocalizationFiles();
+
+	const locales = Object.keys(localizationFiles).filter(
+		locale => locale != 'default'
+	);
+
+	if (locales.length == 0) {
```
jbalsas: This might be a bit strict as a starting point. If someone wanted to support many languages, it would need to manually create as many empty files as necessary. 

Currently, DXP ships with the following settings:
```
locales=ar_SA,eu_ES,bg_BG,ca_AD,ca_ES,zh_CN,zh_TW,hr_HR,cs_CZ,da_DK,nl_NL,nl_BE,en_US,en_GB,en_AU,et_EE,fi_FI,fr_FR,fr_CA,gl_ES,de_DE,el_GR,iw_IL,hi_IN,hu_HU,in_ID,it_IT,ja_JP,kk_KZ,ko_KR,lo_LA,lt_LT,nb_NO,fa_IR,pl_PL,pt_BR,pt_PT,ro_RO,ru_RU,sr_RS,sr_RS_latin,sl_SI,sk_SK,es_ES,sv_SE,th_TH,tr_TR,uk_UA,vi_VN
locales.beta=eu_ES,bg_BG,ca_AD,zh_TW,hr_HR,cs_CZ,da_DK,nl_BE,en_GB,en_AU,et_EE,gl_ES,el_GR,iw_IL,hi_IN,in_ID,it_IT,kk_KZ,ko_KR,lo_LA,lt_LT,nb_NO,fa_IR,pl_PL,pt_PT,ro_RO,ru_RU,sr_RS,sr_RS_latin,sl_SI,sk_SK,tr_TR,uk_UA,vi_VN
locales.enabled=ar_SA,ca_ES,zh_CN,nl_NL,en_US,fi_FI,fr_FR,de_DE,hu_HU,ja_JP,pt_BR,es_ES,sv_SE
```

Either here or in the generator, we should try to:
- Ask for a list of necessary locales
- Generate the locale files
- Save a `locales.enabled` settings or similar that we use to:
    - Create the necessary locale files
    - Translate

izaera: We have the path to the liferay installation if the user provides it. Maybe we can get the list of available locales from there...

izaera: @jbalsas 

I finally discarded the possibility of reading the local DXP installation because it is complex and doesn't add too much value, as the portlet you may be developing can target other platforms.

Also, given that this translate feature needs to have a Microsoft translator key and seems quite specific, I didn't want to force the people using the generator to answer more questions about things they won't probably be using at all.

So, what the generator does now is creating an empty configuration for translation (the `translatorKey` and `supportedLocales` keys of `.npmbuildrc`) and the translate script instructs the user on how to add new supported locales or configure the credentials when it is run.

That way:
1) there's no need to create files manually
2) the script checks for locales which are not supported but have a present .properties file
3) users that don't use the translate feature are not bugged with questions about its configuration

What do you think?

jbalsas:
Sounds good to me! 👍

PR #306, opened 2019-04-05
Fix #305: Fix UnhandledRejectionWarning with EEXIST

### LINKED ISSUE #305: UnhandledRejectionWarning with EEXIST error appears when running bundler
When the bundler is run and a corrupted manifest/build directory is found, the bundler fails with an `UnhandledRejectionWarning` when trying to rename a package directory.

izaera:


packages/liferay-npm-bundler/src/util.js
```diff
@@ -109,6 +101,8 @@ export function renamePkgDirIfPkgJsonChanged(pkg) {
 			getPackageTargetDir(pkgJson.name, pkgJson.version)
 		);
 
+		rimraf.sync(`${newDir}/**`);
+
```
wincent: Changes to `iterateSerially()` look good but how is this related?

wincent: And also, why not `rimraf.sync(newDir)`?

izaera: 1. It is not necessary to fix the bug, but it helped me to diagnose the problem and decided to include it as the code looks much better.

2. :thinking: That's true: `${newDir}/**` should fail because it looks like it won't delete the folder itself but for some reason, it works :man_shrugging: . I'll check again.

izaera: Both seems to work so I'll change it to remove the `**`

PR #329, opened 2019-05-23
Fixes #325: Enhance project detection logic

### LINKED ISSUE #325: Make imports checker stop relying on build.gradle files
Since we can have pure JS projects now, the imports checker should not rely on the existence of build.gradle files.

izaera:


wincent:
LG2M, although I wonder why "node_modules" being present indicates we're in a project dir, but the presence of "package.json" does not...

packages/liferay-npm-imports-checker/src/liferay-npm-imports-checker.js
```diff
@@ -141,6 +140,30 @@ function loadProjects() {
...
 }
 
+/**
+ * Guess if a certain folder contains a project.
+ * @param {string} projectPath
+ */
+function looksLikeProjectDir(projectPath) {
+	if (fs.existsSync(path.join(projectPath, 'build.gradle'))) {
+		return true;
+	}
```
julien: Couldn't we do something like this?

```
function looksLikeProjectDir(projectPath) {
  const filenames = ['build.gradle', '.npmbundlerrc', '.npmbuildrc', 'node_modules'];

  return filenames.some(filename => {
    return fs.existsSync(path.join(projectPath, filename))
  });
}
```

izaera: :+1:

izaera:
> LG2M, although I wonder why "node_modules" being present indicates we're in a project dir, but the presence of "package.json" does not...

In fact, I thought the answer was that we cannot rely on `package.json` because it is present in output folders too and that would fool the tool into thinking that the output of the project is another project.

But that's true for node_modules too although, surprisingly, I have checked this manually and never failed :thinking: .

So I'm going to check again...

PR #337, opened 2019-06-06
Fixes #335: Support for specifying custom manifest headers

### LINKED ISSUE #335: Add support for specifying custom MANIFEST.MF properties
It would be nice to have an extension point to provide custom properties in the MANIFEST.MF file.

Example:

Bundle-SymbolicName: sample-senna-bundle
Bundle-Version: 1.0.0
**Custom-key-version**: 1.0.0-build-14

izaera:


packages/liferay-npm-build-tools-common/src/project/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+import fs from 'fs';
+import path from 'path';
+import readJsonSync from 'read-json-sync';
+
+import Jar from './jar';
+import Localization from './localization';
+
+/**
+ * Describes an standard JS Toolkit project.
```
wincent: "a standard"

packages/liferay-npm-build-tools-common/src/project/localization.js
```diff
@@ -0,0 +1,158 @@
+/**
+ * © 2017 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+import prop from 'dot-prop';
+import fs from 'fs';
+import path from 'path';
+import properties from 'properties';
```
wincent: sort?

packages/liferay-npm-build-tools-common/src/project/index.js
yuchi: in spanish “standard” starts with “e” so requires “an” XD

izaera: :trollface:

packages/liferay-npm-build-tools-common/src/project/localization.js
izaera: It is sorted by module name instead of var name because you can have multiple imports (`import {a,b,z} from 'letters';`) and that would make the sorting algorithm collapse (and probably the Universe the implode).

packages/liferay-npm-build-tools-common/src/project/index.js
izaera: :kissing:

PR #340, opened 2019-06-06
Update package-lock.json files

izaera:


packages/liferay-npm-build-tools-common/src/__tests__/project.test.js
```diff
@@ -60,8 +60,13 @@ describe('standard project', () => {
 	describe('project.jar', () => {
 		it('returns customManifestHeaders', () => {
 			expect(project.jar.customManifestHeaders).toEqual({
-				'Project-Name': 'Test Project',
+				// Coming from .npmbundlerrc
 				Responsible: 'john.doe@somewhere.net',
+				// Coming from manifest.json
+				'Project-Web': 'https://somewhere.net/test-project',
+				'Project-UUID': 'ED7BA470-8E54-465E-825C-99712043E01C',
+				// Coming from both, but .npmbundlerrc wins
```
jbalsas: What's the reason for this order?

packages/liferay-npm-build-tools-common/src/__tests__/project.test.js
```diff
@@ -60,8 +60,13 @@ describe('standard project', () => {
 	describe('project.jar', () => {
 		it('returns customManifestHeaders', () => {
 			expect(project.jar.customManifestHeaders).toEqual({
-				'Project-Name': 'Test Project',
+				// Coming from .npmbundlerrc
```
jbalsas: How is `responsible` coming from `.npmbundlerrc`?

packages/liferay-npm-build-tools-common/src/__tests__/project.test.js
izaera: Which order? I mean, no order exists because it is an object. I just ordered the expectations so that people inspecting the tests may know where each key comes from.

Regarding the "Coming from comments", there are two places for these keys: `.npmbundlerrc` and `manifest.json`.  In general you would use one of the other, but if both are used we need to disambiguate and I chose the completely arbitrary precedence order:  `.npmbundlerrc` > `manifest.json`.

I say it is arbitrary, but looks more logic to me because `.npmbundlerrc` is  more low level, so it makes sense that values there override those in the higher level `manifest.json` file...

packages/liferay-npm-build-tools-common/src/__tests__/project.test.js
izaera: See previous conversation.

packages/liferay-npm-build-tools-common/src/__tests__/project.test.js
jbalsas: I think it is exactly the opposite. Higher level should trump lower level, because it's where you have the most information about the project. 

If IT sets up the build process (or we do it), we don't have all the information to make the decision. I think `manifest.json` should win over `.npmbundlerrc`

packages/liferay-npm-build-tools-common/src/__tests__/project.test.js
jbalsas: See previous answer, but still, where is `responsible` coming from? I assume from your answer it's not an out of the box key we add, but something in this test in particular?

packages/liferay-npm-build-tools-common/src/__tests__/project.test.js
izaera: OK. I'll reverse it then.

packages/liferay-npm-build-tools-common/src/__tests__/project.test.js
izaera: It's in the test fixtures -> https://github.com/liferay/liferay-js-toolkit/blob/3180f61b68d95a620ac7afb23bbd0507f36d511f/packages/liferay-npm-build-tools-common/src/__tests__/__fixtures__/project/standard/.npmbundlerrc

We don't add any out of the box keys through .npmbundlerrc. I think that explains your previous comment that I didn't quite understood :sweat_smile: :

> If IT sets up the build process (or we do it) ...

We don't set anything in `.npmbundlerrc` normally. Any key we add (like Resource-Bundle, Tool, or Require, ...) that is needed by the JAR to work correctly is hard coded into the bundler, not taken from the `.npmbundlerrc`. 

`.npmbundlerrc` and `manifest.json` are for the user only. At least for now...

izaera: But yes, it may make sense in some scenarios that someone adds a key through a preset in liferay-npm-bundler and then someone else overrides it in a manifest.json.

Issue #328, opened 2019-05-23
Create an abstraction library to reflect on projects' configuration


izaera:
We should extract a common library to read .npmbundlerrc configuration as well as .npmbuildrc or any other project configuration file. It's the only way to make sure that every tool in the JS toolkit "sees" the same things.

In other words, the project's model should be abstracted by a library instead of directly read from configuration files by each tool.

izaera:
This is not yet finished

izaera:
This will be totally fixed after #429

PR #346, opened 2019-06-10
Fixes #345 ("Custom filename for generated jar")

### LINKED ISSUE #345: Custom filename for generated jar
As I am working through the development/deployment lifecycle for a new portal project, it seems desirable to control the filename of the generated jar (specifically to remove the last hyphen, as well as the version number, from the generate filename when using the "create-jar" option in .npmbundlerrc), similar to how the output directory is both defaulted and configurable.

(pull request coming shortly, as I believe that I had worked out how to do this locally - although I realize that some additional documentation changes would likely be required, and am entirely open to having glossed over an existing way to accomplish this same bit).

reinrl:
Maintains the existing default ('${pkgJson.name}-${pkgJson.version}.jar') jar filename, while also allowing for the optional ability to provide a static string filename within .npmbundlerrc.

### DIFF
--- a/packages/liferay-npm-bundler/src/config/internal/jar.js
+++ b/packages/liferay-npm-bundler/src/config/internal/jar.js
@@ -95,6 +95,18 @@ export function getOutputDir() {
 }
 
 /**
+ * Get the output file name for JAR files. Defaults to an empty string if none is
+ * specified, requiring the caller to generate the default based upon the package 
+ * name and version specified in the package.json.
+ * @return {string}
+ */
+export function getOutputFilename() {
+	const jarConfig = getNormalizedJarConfig();
+
+	return prop.get(jarConfig, 'output-filename', '');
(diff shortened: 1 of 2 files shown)


izaera:
Just started reviewing :)

izaera:
@reinrl The contribution is OK, but this change has been sent in the middle of #328 (I'm doing heavy refactors for that issue :sweat_smile: ) so I will change the location of your code to `liferay-npm-build-support`'s `Project` class, which will be the final place to centralize all config handling code.

I'll resend this PR with the modified code :+1: .

izaera:
I was referring to these changes :point_right:  https://github.com/izaera/liferay-js-toolkit/commits/issue-347 in the previous comment

izaera:
Resent as https://github.com/liferay/liferay-js-toolkit/pull/348

PR #348, opened 2019-06-12
Fixes #345: Move default logic to config side

### LINKED ISSUE #345: Custom filename for generated jar

izaera:


izaera:
Resending https://github.com/liferay/liferay-js-toolkit/pull/346 with some minor changes.

PR #351, opened 2019-06-19
Fixes #347: Add support for create-react-app bundling

### LINKED ISSUE #347: Add support for create-react-app bundling
Based on [this spike](https://issues.liferay.com/browse/IFI-734) we will proceed with supporting bundling of create-react-app projects.

izaera:


packages/liferay-npm-bundler-plugin-create-react-app/src/index.js
```diff
@@ -29,4 +29,7 @@ export default function({config, log, pkg, rootPkgJson, source}, {pkgJson}) {
 	pkgJson.portlet['com.liferay.portlet.header-portlet-css'] = cssFiles.map(
 		cssFile => `react-app/static/css/${path.basename(cssFile)}`
 	);
+
+	// Add SPA off option to portlet
+	pkgJson.portlet['com.liferay.portlet.single-page-application'] = false;
```
matuzalemsteles: It's worth pointing out a negative point when disabling the SPA here, let me know if I'm wrong about this, but when the user is interacting with the portlet with a React Router for example, when the user exits from the portlet control page to another page of Liferay there will be navigation made by default browsers without control of a SPA, when the user tries to return to the previous page it comes back again with the standards browsers. In this case we lose the benefit of trying to create an integration to create smooth transitions between navigations.

I had done a POC (https://github.com/jbalsas/liferay-portal/pull/1687) that only made a few adjustments, where I was dealing with this problem (there is still a lot of room to be improved there.) The SPA lost the navigation control when it entered a portlet with react-router for example, but it was back in action when the user clicked on another liferay-portal menu. The benefit of this is that if we have a large navigation history between exits and entries between portlets that navigation is done by SPA and React Router, transitions to the user will seem smooth.

yuchi: @matuzalemsteles Missing support for client-side routers is a big pain when developing for Liferay DXP. My initial suggestion was described in liferay/senna.js#180 and should cover most cases by providing specialized `history` objects to React Router and similar libraries.

Please `@mention` me in future discussion about this.
I’d include @faragos also since we just had a similar discussion on the Liferay Community chat few days ago.

yuchi: It’s also worth noting the fact that Senna is more or less useless in the DXP scenario. Most of its code is meant to be used with different surfaces, but in DXP there’s only one…

AFAIL @zenorocha and @eduardolundgren initially implemented Senna with the concept of surfaces in mind so that they could be used to map portlets. But themes and different page layouts made it non viable.

matuzalemsteles: @yuchi yeah, I agree with you about this, this is really weird, I do not know the real reason why they went with this to the DXP. But I want to try taking senna just to handle these integration cases and handle those transitions until we can switch to another more robust router with the technology we're choosing...

izaera: @matuzalemsteles :thinking: When I did that I was thinking about making it configurable but then decided to hard code it and then change it if it was necessary (mostly because I didn't know the full consequences :sweat_smile: ).

Do you think making this configurable would fix the issue?

The thing is that we must have always in mind that the model is that the user gets an existing `create-react-app` application and deploys it to Liferay. He's not doing a Liferay specific application because in that case it would -probably- make more sense to use the alternative stack (bundler + imports and forget about create-react-app adaptation). 

I'm aware that the previous reasoning is not 100% true, but I hope you get the idea. We are trying to make it easier to deploy an agnostic React application to Liferay. Much in the same way you would make an application that targets, say, the browser and Cordova at the same time.

It's clear that such type of application could have platform specific code. For example, a typical:

```javascript
if (window.cordova) {
  // call Cordova API
} else {
  // emulate Cordova API or tell the user that's not supported in a webapp
}
```

But that's very far from making the whole application routing dependant on Liferay's specific mechanisms, IMO. :thinking: 

Thoughts?

izaera: OK, talking with @jbalsas , I now realise that all this discussion has no much sense :sweat_smile: .

The `create-react-app`/`webpack` adaptation only makes sense for single portlets that render the whole page. You cannot (at least as it is now) place more than one webpack adapted portlet per page.

I know you could still argue that there's a use case where you have just one webpack based portlet in a page and the rest are standard portlets, but it's a degenerated case IMO which is not worth covering (at least for now).

izaera:
I'm resending this as https://github.com/liferay/liferay-js-toolkit/pull/376 due to rebase conflict.

Issue #303, opened 2019-04-04
Add support for static resource loaders (treat CSS, SVG, JSON, etc. as JS modules that may be imported)


rozamirandar:
It would be nice to have a way of dynamically loading CSS code as it's done with Javascript AMD Modules.

@izaera @jbalsas

sohui:
It would be great to have this feature.

sohui:
`import styles from './index.css';
import { formatMessage } from 'umi-plugin-locale';

export default function() {
  return (
    <div className={styles.normal}>
      <div className={styles.welcome} />
      <ul className={styles.list}>
        <li>To get started, edit <code>src/pages/index.js</code> and save to reload.</li>
        <li>
          <a href="https://umijs.org/guide/getting-started.html">
            {formatMessage({ id: 'index.start' })}
          </a>
        </li>
      </ul>
    </div>
  );
}
`
when import css, it raised errors.

ERROR in ../src/pages/index.css 2:0
Module parse failed: Unexpected token (2:0)
You may need an appropriate loader to handle this file type.

izaera:
@rozamirandar In addition to CSS resources, webpack also supports static media (like svg, jpg, and the like).

The general contract seems to be:

1. `import css` adds the CSS to the CSS bundle (in our case it should add a `com.liferay.portlet.header-portlet-css` property to the portlet, probably)
2. `import svg|jpg|...` returns a string with the URL of the resource (in our case we should look at the web context path to know how to get it, but I'm not sure what to do with the `/o` prefix: should we hard code it? make it configurable just in case someone changes it? ...?)

izaera:
Also, we need to take into account that the portal may have a proxy, CDN, or a custom webapp root configured. I don't know if that needs anything specific whether in the server or the build side... :thinking:

jbalsas:
I think we'll need different CSS loaders... if someone's doing some type of CSS-in-JS, they might just want a representation of the CSS stylesheet (maybe a `CSSStyleSheet`?). Someone else might want to just throw that into a `style` tag. 

Similarly for svgs, I've seen svg-loaders to actually expose the different symbols inside the svg.

Ideallly, we would want some type of glob-to-loader mechanism.

izaera:
:thinking: Making the assumption that this kind of imports only appear in the project's code (as opposed to third party dependencies), the easiest ways to tackle this is -probably- to use babel plugins during the bundler's build phase to transform these `require`s into something meaningful.

[...]

PR #356, opened 2019-06-24
Add loaders to npm bundler build process

rozamirandar:
@jbalsas 

Issue: https://github.com/liferay/liferay-js-toolkit/issues/303

This is a possible implementation to support loading CSS code via AMD. I tried to get as close as possible to what exists in Webpack CSS Loaders.

The **liferay-npm-bundler-loader-css-loader** receives the CSS file content and generates Javascript code that later will be transformed into an AMD module by babel and liferay-npm-bundler plugins. Since the CSS file is transformed into a Javascript AMD module, if some Javascript module declare it as a dependency via `import './Style.css'`, the Liferay AMD Loader will be able to load it at runtime.

This CSS loader simply creates a style tag during runtime and pushes the CSS content inside it and append it to the `<head>` tag, new features and options can be added if requested.

In order to test it please take a look at the css-loader README file.
Change the css-loader dependency to your local directory in the file `packages/liferay-npm-bundler-preset-standard/package.json`


phillipsheridan:
Let's say we have a widget page with a react portlet. That portlet generates style tags inside the head tag of the page. Now we navigate to another widget page, which also adds more style tags to the head tag. The style tags from the first portlet never leave the head tag. How do we stop the style tags from one portlet leaking into other portlets? This is a big problem we are facing right now. 

I think that it may be beneficial to find a way to load all these style tag's selectors into the portlet's inner css file so the styles are not lingering in the head tag.

phillipsheridan:
It's super hacky, but I took care of this by doing something like the following in the javascript page configuration that runs on each page:

```
var purgeStyles = function() {
  var hs = document.getElementsByTagName('style');
  for (var i = 0, max = hs.length; i < max; i++) {
    // console.log('length is ', max);
    var child = hs[i];
    var parent = hs[i].parentNode;
    if (parent.nodeName === 'HEAD') {
      parent.removeChild(child);
      max--;
      i--;
    }
  }
};

$(document).on(
  'click',
  `a[href^="${window.location.origin}/group/guest"]`,
  purgeStyles
);
```

jbalsas:
Hey @phillipsheridan, on the Liferay Portal side, Senna should be taking care of this on navigation. If the widget is no longer in the page, it should be removing its related resources. Could you try annotating your styles with the `data-senna-track="temporary"` attribute as explained in https://sennajs.com/docs#trackableresources?

phillipsheridan:
@jbalsas The page you linked doesn't seem to mention anything about style elements? How would I annotate those?

rozamirandar:
@jbalsas @phillipsheridan I'll be adding the senna tag to the loaders

phillipsheridan:
@rozamirandar Yes that would help. In our current solution I just added
```
{
            loader: 'style-loader',
            options: { attrs: { 'data-senna-track': 'temporary' } }
          },
          'css-loader',
          'sass-loader'
        ]
      }
```
 But it doesn't completely solve the problem. Currently we are using Material UI for react. Behind the scenes, MUI creates a bunch of style tags.



And it seems other libraries do this as well. The last one with the data-emotion attr is from another popular library, Emotion. We have no control over those style elements it seems. I'm researching a possible solution at the moment.

izaera:
Just started reviewing :)

izaera:
@rozamirandar I'm reviewing the PR. Do we really need the new notion of "loaders" in the bundler? I think this could be solved with plain babel or bundler plugins (or a combination of both...) :thinking:

jbalsas:
Hey @izaera, that's a legitimate question :)

I think I might've been the one to steer @rozamirandar towards this notion, but you're right that if we can achieve this with some other construct already present that would be enough too.

The only part that might be missing is that you usually can provide a matching function/regexp for loaders, to specify which files need to go through which transforms. I'm not sure how can we exactly get that behaviour without some custom logic...

izaera:
@jbalsas I'm not against this "new" addition, but it's a paradigm shift in the sense that we are moving part of the build to the bundler, similar to what webpack does. Apparently the loaders are generating output code which is something we have always done before the bundler kicks in.

I have considered the possibility to put the bundler in charge of the full build previously (as webpack does), because it has a lot of benefits, but we need to design a proper roadmap for that, otherwise we may end up with two configuration/execution models in a single tool.

izaera:
@rozamirandar After talking with @jbalsas in a private channel, we think that it makes sense to explore the loader plugins concept so that we can start a paradigm shift to put the bundler as the main driver of project builds, instead of making it appear as a last step.

So, you don't need to reply to my previous question :sweat_smile:

izaera:
One thing this PR is missing is the possibility to run loaders on third party packages (those living in node_modules as opposed to the ones in the project itself). 

Webpack doesn't treat third party modules different from project modules.

I'm going to do the same as webpack and run loaders for third party projects too. If anyone sees any possible problem please raise your hand :-).

izaera:
Resent as https://github.com/liferay/liferay-js-toolkit/pull/378

Thx for your great work and inspiration @rozamirandar :-)

MarcoPortillo:
Hi, 

I am still having problems with this, I cannot load any CSS files from npm packages (node_modules) there are many libraries such as https://github.com/dangrossman/daterangepicker I want to use, but I can't use due to loaders not working, I can only use my own css files if I import/include them in /css/styles.css, otherwise doesn't work. I tried with this https://github.com/liferay/liferay-js-toolkit/wiki/List-of-loaders is still not working

izaera:
@MarcoPortillo what did you try exactly? We need something to reproduce the problem or analyze what's wrong.

You may start opening a new issue and posting the liferay-npm-bundler-report.html file after cleaning and building your project (see [this option in npmbundlerrc](https://github.com/liferay/liferay-js-toolkit/wiki/.npmbundlerrc-file-reference#dump-report) to turn reports on)

PR #357, opened 2019-06-26
React portlet update

phillipsheridan:
I made these changes to take care of a few issues I found with the React portlet generator:

1. I think that Sass should be the default for development. I tried just using the standard sass webpack loaders for inline styles but since webpack is not used for the build process, that wasn't going to work. So I added node-sass to watch the assets folder. For node sass to work on windows, node-gyp is needed. Node-gyp needs windows-build-tools so I added that. Currently they still get installed on unix machines too and I haven't tested that. I wasn't sure of a way to check OS.

2. Class properties are currently stage 3. It's pesky writing constructors just to initialize state and binding functions. So I updated the babel presets and added the class-properties plugin for modern react features.

 I am open to discussion of optimizing the new features.

### DIFF
--- a/packages/babel-plugin-name-amd-modules/package-lock.json
+++ b/packages/babel-plugin-name-amd-modules/package-lock.json
@@ -862,7 +862,8 @@
 				},
 				"buffer-shims": {
 					"version": "1.0.0",
-					"bundled": true
+					"bundled": true,
+					"optional": true
 				},
 				"caseless": {
 					"version": "0.12.0",
@@ -893,11 +894,13 @@
 				},
 				"console-control-strings": {
 					"version": "1.1.0",
-					"bundled": true
+					"bundled": true,
+					"optional": true
 				},
 				"core-util-is": {
 					"version": "1.0.2",
-					"bundled": true
+					"bundled": true,
+					"optional": true
 				},
 				"cryptiles": {
 					"version": "2.0.5",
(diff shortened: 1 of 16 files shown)


jbalsas:
Hey @phillipsheridan, thanks a lot for sending this over!

We currently have some work in progress related to this from @rozamirandar at https://github.com/liferay/liferay-js-toolkit/pull/356. 

I'd like to take a look at both these PRs together to make sense of them as a whole. Could you take a look and see if that generic approach could serve your most pressing needs here as well or if we could shape it in a way that it would?

@rozamirandar, could you also take a look at this and see if this and yours are compatible or in fact orthogonal and independent?

I'm currently in between PTO and @izaera is too. We'll be back full strength on monday and will be looking to prioritize this.

Thanks a lot for your patience!

rozamirandar:
@jbalsas from what I could see this is independent of what I've done.

jbalsas:
@rozamirandar, wouldn't a general sass-loader cover that use case? Currently we created a css one, but certainly we will need one to compile sass and wrap the result in a consumable module.

rozamirandar:
@jbalsas Yes, I think so. I already added the sass-loader and styles-loader to my PR.

phillipsheridan:
Hey guys I'm gonna either edit this pull request if it is possible or create a new one to only include the babel updates since @rozamirandar handled the styles.

phillipsheridan:
closed so I can open a new one

PR #361, opened 2019-07-10
React portlet update v2

phillipsheridan:
Updated babel & added support for class properties

### DIFF
--- a/packages/babel-plugin-name-amd-modules/package-lock.json
+++ b/packages/babel-plugin-name-amd-modules/package-lock.json
@@ -862,7 +862,8 @@
 				},
 				"buffer-shims": {
 					"version": "1.0.0",
-					"bundled": true
+					"bundled": true,
+					"optional": true
 				},
 				"caseless": {
 					"version": "0.12.0",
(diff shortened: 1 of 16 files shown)


izaera:
Hi @phillipsheridan . Sorry for the delay, but merging in the PR for #303 is proving to be a tough one :sweat: .

Regarding this PR, I like the idea of moving to babel 7 and that's in fact something that we wanted to do but didn't tackle it because it needs to be done for all project types, not just React. If you want to try it in all projects you are welcome ;-) . If not, I will get your commit and replicate it in the other project types as well, some time in the future.

There's also a modification to add `@babel/plugin-proposal-class-properties` to the build, but that wouldn't be merged. The rationale behind it is:

1) It is a proposal, so it's not clear that it's so general that everyone will want it
2) It's something you may easily add to your project once it is generated

In general, we don't add support for anything in our generated projects that it is not so indisputably mainstream that nobody will want to remove it. So, we don't aim to having full-featured projects generated by our yeoman generator, but to having reasonable projects that can be easily extended to tailor your needs. 

Keep in mind that what may seem reasonable to you may be something useless or forbidden in some other place. Also, even if the JS toolkit wants to make our lives easier, it doesn't want to be opinionated on how you should structure your project.

izaera:
I'm closing this because once we release [bundler 3](https://github.com/liferay/liferay-js-toolkit/milestone/52), everybody will be able to use whatever they want to transpile their projects and we'll use babel 7 for generated projects.

Bundler 2 will probably remain as it is (using babel 6) unless there's a big demand for the upgrade by users that can't switch to bundler 3.

Thx for the contribution, anyway :-). And sorry because it didn't get merged in the end :-(.

PR #372, opened 2019-07-26
Fixes #366: If create-jar.output-filename is set, deploy should use that as the jar filename

### LINKED ISSUE #366: If create-jar.output-filename is set, deploy should use that as the jar filename

izaera:


izaera:
Resent to the correct branch as https://github.com/liferay/liferay-js-toolkit/pull/373

PR #376, opened 2019-08-05
Fixes #347: Add support for create-react-app bundling 

### LINKED ISSUE #347: Add support for create-react-app bundling

izaera:


packages/liferay-npm-build-support/src/scripts/build/create-react-app.js
```diff
@@ -0,0 +1,479 @@
...
+						return;
+					}
+
+					getElementByIdCallArgs[0] = t.identifier(
+						'portletElementId'
+					);
+
+					if (state.renderFound) {
+						throw Object.assign(new Error(), {
+							humanMessage: msg.noValidEntryPoint,
```
wincent: I found this a bit confusing on first reading it, because I thought, "why are we using the same error message for `state.renderFound` here and `!state.renderFound` below?". Then I read the error message text and saw that it wants exactly 1 render call.

I think you could avoid that by changing the `state.renderFound` (boolean) to `state.renderCount` (number). So here you would replace the `if`/`else` with:

```javascript
state.renderCount++;
```

and below on line 423 you'd replace the `if (!state.renderFound)` conditional with:

```javascript
if (state.renderCount !== 1) {
```

packages/liferay-npm-build-support/src/util.js
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+/**
+ *
+ * @param {string} script
+ * @param {Array<*>} args
+ */
+export function runNpmBin(script, args = []) {
+	child_process.spawnSync(
```
wincent: Should we use `cross-spawn` to avoid possible issues on Windows? (I have had to fix some bugs like that in our other projects.)

packages/liferay-npm-build-tools-common/src/format.js
```diff
@@ -0,0 +1,143 @@
...
+ */
+export function print(lines) {
+	if (!Array.isArray(lines)) {
+		lines = [lines];
+	}
+
+	lines = lines.map(line => {
+		line = removeJsFormatWhitespace(line);
+
+		line = line.replace(/(https?:\/\/[^\s]+)/gm, fmt.underline('$1'));
```
wincent: So this could underline too much:

```
Some link (https://example.com/foo).
           ~~~~~~~~~~~~~~~~~~~~~~~~~
```

packages/liferay-npm-build-tools-common/src/format.js
```diff
@@ -0,0 +1,143 @@
...
+
+	console.log(lines.join('\n'));
+}
+
+export function success(literals, ...values) {
+	return fmt.success(format('✔️', literals, values));
+}
+
+export function question(literals, ...values) {
+	return fmt.question(format('🙋', literals, values));
```
wincent: Possible gender issue with using a woman could be taken to insinuate that women are not experts.

❓

packages/liferay-npm-build-tools-common/src/format.js
```diff
@@ -0,0 +1,143 @@
...
+ * This function is (and must always be) idempotent because it may need to be
+ * applied several times to the same line.
+ * @param {string} line
+ */
+function removeJsFormatWhitespace(line) {
+	line = line.replace(/^\n/, '');
+	line = line.replace(/^\t\t/gm, '');
+
+	return line;
+}
```
wincent: The hard-coding of the leading two-tabs is a bit brittle. I have had to write a few of these `dedent` functions over the years; [here is an example](https://github.com/wincent/js/blob/5d8e5e803f7fd30c2c12d2497c6d38c593a95752/packages/dedent/src/index.ts).

packages/liferay-npm-build-tools-common/src/project/index.js
```diff
@@ -58,6 +60,51 @@ export class Project {
...
+			? readJsonSync(npmbundlerrcPath)
+			: {};
+
+		// Apply preset if necessary
+		let presetFile;
+
+		if (config.preset === undefined) {
+			presetFile = require.resolve('liferay-npm-bundler-preset-standard');
+		} else if (config.preset === '' || config.preset === false) {
+			// don't load preset
```
wincent: What about `null`? Might that be passed?

packages/liferay-npm-bundler-plugin-create-react-app/src/index.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * @return {void}
+ */
+export default function({pkg}, {pkgJson}) {
+	// NOTE: The name of the "main" file must be in synch with the one generated
```
wincent: synch -> sync

packages/generator-liferay-js/src/adapt/index.js
```diff
@@ -0,0 +1,260 @@
...
+		{'--force'} argument to Yeoman.
+		`,
+	],
+	welcome: [
+		title`
+		
+		|👋 |Welcome to Liferay JS Toolkit project adapter 
+		`,
+	],
+};
```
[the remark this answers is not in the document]
izaera: :heart_eyes:

packages/generator-liferay-js/src/utils/JsonModifier.js
```diff
@@ -37,7 +40,7 @@ export default class JsonModifier {
 
 		modifier(json);
 
-		gen.fs.write(this._path, JSON.stringify(json, null, '	'));
+		gen.fs.write(this._path, JSON.stringify(json, null, this._space));
```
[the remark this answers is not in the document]
izaera: :heart_eyes:

packages/generator-liferay-js/src/utils/index.js
```diff
@@ -184,3 +220,20 @@ export function toHumanReadable(string) {
 
 	return humanizedString;
 }
+
+/**
+ * Check if a given directory path contains a valid Liferay installation.
+ * @param  {String} input directory path
+ * @return {boolean}
```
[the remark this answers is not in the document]
izaera: Yes, it's an API :man_shrugging: .

packages/generator-liferay-js/src/utils/modifier/package.json.js
```diff
@@ -14,9 +14,11 @@ import JsonModifier from '../JsonModifier';
 export default class extends JsonModifier {
 	/**
 	 * @param {Generator} generator a Yeoman generator
+	 * @param {string|number} space the space string/number of spaces to use
+	 * 			when stringifying
 	 */
-	constructor(generator) {
-		super(generator, 'package.json');
+	constructor(generator, space = '	') {
```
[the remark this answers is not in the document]
izaera: :open_mouth:

packages/generator-liferay-js/src/utils/modifier/npmbundlerrc.js
```diff
@@ -66,4 +66,14 @@ export default class extends JsonModifier {
...
 		});
 	}
+
+	/**
+	 * Set preset package name.
+	 * @param {string} presetName
+	 */
+	setPreset(presetName) {
+		this.modifyJson(json => {
+			prop.set(json, `preset`, presetName);
```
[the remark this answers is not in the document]
izaera: So that you can offer me some feedback :smiling_imp: .

packages/liferay-npm-build-support/src/resources/build/create-react-app/index.js.ejs
```diff
@@ -0,0 +1,60 @@
...
+      (function() {
+        var entryPoint = undefined;
+        var pendingLaunchs = [];
+
+        return {
+          launch: function(args) {
+            if (entryPoint) {
+              entryPoint(args);
+            }
+            else {
```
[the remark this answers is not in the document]
izaera: :+1:  I'm changing it. However it doesn't matter because it ends up inside a JAR file, not in a buildable artifact so we can consider it as derived output.

packages/liferay-npm-build-tools-common/src/format.js
izaera: And not using a woman could be taken to insinuate that we only think in men :man_facepalming: :woman_facepalming: . I usally use gender icons interchangeably :woman_shrugging: .

izaera: I tried the question icon first but for some reason changed it. Can't remember why. I will give it a try.

packages/liferay-npm-build-tools-common/src/format.js
izaera: I didn't want to complicate it until we really need it, so I decided to do it as simple as possible. If you follow the rules (message tables and formatting rules), this should work for every situation.

I really think the tables are useful, because they remove noise from source code and make the tools I18N-able which is something we may need in the future.

packages/liferay-npm-build-tools-common/src/project/index.js
izaera: Nope, because it comes from a JSON file.

packages/liferay-npm-build-tools-common/src/format.js
izaera: :question: seems to work. I will use it.

packages/liferay-npm-build-tools-common/src/format.js
izaera: It's complex, because `)` and `.` are legal URL characters so there's no easy way to know that they are not part of the URL (for a computer, I mean). 

I would say that in that case, we may add a space after the URL, for example. Or, when the case comes, create a new syntax to explicitly delimit the URL, for example.

packages/liferay-npm-build-support/src/util.js
izaera: Good catch.

packages/liferay-npm-build-tools-common/src/project/index.js
wincent: But `null` is representable in JSON.

packages/generator-liferay-js/src/adapt/index.js
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	/**
+	 * Standard Yeoman install function
+	 */
+	install() {
+		print(msg.projectAdapted);
+
+		this.installDependencies({
+			bower: false,
```
jbalsas: If I'm not mistaken, `create-react-app` uses `yarn`,  so we should too.

```suggestion
			bower: false,
			npm: false,
			yarn: true,
```

jbalsas: See https://yeoman.io/authoring/dependencies.html for reference

PR #378, opened 2019-08-13
Fixes #303: Support for loaders

### LINKED ISSUE #303: Add support for static resource loaders (treat CSS, SVG, JSON, etc. as JS modules that may be imported)

izaera:


packages/liferay-npm-bundler/src/config/internal/bundler.js
```diff
@@ -81,6 +81,47 @@ export function getPlugins(phase, pkg) {
...
+
+	// Normalize rules
+	rules.forEach(rule => {
+		// `rule.files` must be array
+		if (typeof rule.files === 'string') {
+			rule.files = [rule.files];
+		}
+
+		// `rule.use` must be an array
+		if (!Array.isArray(rule.use)) {
```
wincent: Why did you perform the "must be array" check in two ways (once with `typeof ... === 'string'` and once with `!Array.isArray()`)?

packages/liferay-npm-bundler/src/config/internal/bundler.js
```diff
@@ -122,3 +163,37 @@ function instantiatePlugins(pluginNames) {
...
+					`Loader '${resolvedLoader}' is incorrect: ` +
+						`it does not export a function`
+				);
+			}
+
+			return {
+				loader,
+				options,
+				resolvedLoader,
+				exec,
```
wincent: You know you want to alphabetically sort this.

packages/liferay-npm-bundler/src/report/html.js
```diff
@@ -102,6 +104,10 @@ export function htmlDump(report) {
 						copyKeys.length > 0,
 						() => `${copyKeys.length} plugins applied`
 					);
+					const rulesNotice = htmlIf(
+						rulesKeys.length > 0,
+						() => `${rulesKeys.length} files processed`
```
wincent: nit: if there were only 1 we'd still emit `1 files` (although I know that's already true elsewhere in this file).

packages/liferay-npm-bundler/src/runners.js
```diff
@@ -216,6 +225,67 @@ function processBabelFiles(filePaths, chunkIndex, pkg, babelConfig) {
...
+
+	const filePromises = globby.sync(globs).map(filePath => {
+		const content = fs.readFileSync(filePath, 'utf-8').toString();
+		const log = new PluginLogger();
+
+		const context = {
+			content,
+			filePath,
+			extraArtifacts: {},
+			log,
```
wincent: Sort?

resources/devtools/link-js-toolkit/link-js-toolkit.js
```diff
@@ -14,12 +14,29 @@ const {safeRunFs, yarn} = require('./util');
 // Read package.json
 const pkgJson = readJsonSync(path.join('.', 'package.json'));
 
-// Grab dependencies
-const deps = []
+let deps;
+
+// Grab dependencies from package.json
+deps = []
```
wincent: You could start this chain with the first `Object.keys()` result rather than creating an empty array, but would probably look nicer with spread syntax anyway:

```javascript
let deps = [
    ...Object.keys(pkgJson.dependencies),
    ...Object.keys(pkgJson.devDependencies),
].filter(...);
```

resources/devtools/link-js-toolkit/link-js-toolkit.js
```diff
@@ -14,12 +14,29 @@ const {safeRunFs, yarn} = require('./util');
...
+deps.push(...dirs);
+
+// Deduplicate dependencies
+deps = Object.keys(
+	deps.reduce((map, item) => {
+		map[item] = item;
+		return map;
+	}, {})
+);
+
```
wincent: Simpler I think is to use a `Set`:

```javascript
deps = Array.from(new Set(deps));
```

packages/liferay-npm-bundler-loader-babel-loader/README.md
```diff
@@ -0,0 +1,38 @@
...
+npm install --save-dev liferay-npm-bundler-loader-babel-loader
+```
+
+## Usage
+
+In order to use this loader you must declare a rule in your module's `.npmbundlerrc` file:
+
+```json
+{
+	"packages": {
```
jbalsas: As discussed offline, this is the main thing that I don't like...

I'd like to focus this on the loaders and not on the packages so it's clearer what you're configuring.

I imagine something like:

```javascript
{
    "rules": [
        {
            "..."
        }
    ]
}
```

Specificity and package-scope should ideally be handled within the `test`/`files` configuration of the loader.

/@wincent, thoughts?

wincent: That sounds ok to me, if I am understanding you correctly (for me the `.npmbundlerrc` file is kind of like configuring webpack, which means I don't understand it unless I have the documentation open right next to me 😂).

mwilliams2014:
@izaera please let me know when this is complete, so we can draft documentation for it

packages/liferay-npm-bundler-loader-babel-loader/README.md
izaera: This change has been pushed now. Resolving.

packages/liferay-npm-bundler-loader-babel-loader/src/__tests__/index.test.js
```diff
@@ -0,0 +1,49 @@
...
+	const context = {
+		content: 'var x = 1;',
+		filePath: 'file.js',
+		log: new PluginLogger(),
+		extraArtifacts: {},
+	};
+
+	loader(context, {});
+
+	expect(context.log.messages).toMatchSnapshot();
```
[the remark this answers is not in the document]
izaera: I sometimes use snapshots to make code cleaner or when the expected results is big and difficult to type/understand when written in the test directly. 

That, of course, is a matter of personal taste more than a clearly defined rule so it may/may not apply in this case.

packages/liferay-npm-bundler-loader-babel-loader/src/index.js
```diff
@@ -0,0 +1,106 @@
...
+		babelConfig.plugins || []
+	);
+	delete babelConfig.presets;
+
+	const result = babel.transform(content, babelConfig);
+
+	context.sourceMap = result.map;
+	context.extraArtifacts[`${filePath}.map`] = JSON.stringify(result.map);
+
+	log.info('babel-loader', 'Transpiled file');
```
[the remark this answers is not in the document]
izaera: There's no need to include it because each `log` object is associated to a single file so that when it is dumped to the report file you see the name of the file heading the contents of the `log`.

packages/liferay-npm-bundler-loader-babel-loader/src/index.js
```diff
@@ -0,0 +1,106 @@
...
+			...presets.map(preset => {
+				let presetModule, presetName;
+
+				try {
+					presetName = `babel-preset-${preset}`;
+					presetModule = projectRequire(presetName);
+				} catch (err) {
+					presetName = preset;
+					presetModule = projectRequire(presetName);
+				}
```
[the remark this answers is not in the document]
izaera: :thinking: Probably. And maybe we should get `babel` from the project too (instead of always using the one we bundle with the loader).

@jbalsas Thoughts?

packages/liferay-npm-bundler-loader-babel-loader/src/index.js
```diff
@@ -0,0 +1,106 @@
...
+
+				if (presetModule === undefined) {
+					throw new Error(
+						`Babel preset '${presetName}' does not export anything`
+					);
+				}
+
+				if (presetModule)
```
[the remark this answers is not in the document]
izaera: I'm removing the if. It seems to be a leftover of some refactor or rebase :-(.

Regarding the false-y check... Yes, you could make it that way too but I want to throw an error when a module does not export anything, not when it exports something incorrect. That's why I add the other checks below.

packages/liferay-npm-bundler-loader-style-loader/src/index.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @param {object} context loader's context
+ * @return {string} the processed file content
+ */
+export default function(context) {
+	const {filePath, log} = context;
+	let {content} = context;
+
+	content = content.replace(/\n/g, '\\n');
+	content = content.replace(/"/g, '\\"');
```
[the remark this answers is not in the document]
izaera: Nice catch :raised_hands:

packages/liferay-npm-bundler/src/config/internal/bundler.js
izaera: This has moved to other file, so it's not longer valid.

packages/liferay-npm-bundler/src/config/internal/bundler.js
izaera: This has moved to other file, so it's not longer valid.

packages/liferay-npm-bundler/src/runners.js
izaera: This has moved to other file, so it's not longer valid.

PR #380, opened 2019-08-13
Fixes #379: Perform simple dead code elimination as part of bundling

### LINKED ISSUE #379: Perform simple dead code elimination as part of bundling
Related to: https://issues.liferay.com/browse/LPS-95430

React has special code to detect when it is run in production mode but dead code elimination has not been performed ([see here](https://github.com/facebook/react/blob/baff5cc2f69d30589a5dc65b089e47765437294b/packages/react-dom/npm/index.js#L3)). This triggers an ugly error in the console:

> Uncaught Error: React is running in production mode, but dead code elimination has not been applied. Read how to correctly configure React for production: ...

In order to avoid this kind of error, and additionally avoid bundling development code in production (or production code in development), we can use a minimal and conservative form of dead code removal, such as that provided by [babel-plugin-minify-dead-code-elimination](https://www.npmjs.com/package/babel-plugin-minify-dead-code-elimination).

wincent:
Related to: https://issues.liferay.com/browse/LPS-95430

React has special code to detect when it is run in production mode but dead code elimination has not been performed (see [here](https://github.com/facebook/react/blob/baff5cc2f69d30589a5dc65b089e47765437294b/packages/react-dom/npm/index.js#L3)).

This ends up throwing an error:

> Uncaught Error: React is running in production mode, but dead code
> elimination has not been applied. Read how to correctly configure
> React for production: ...

In order to avoid this kind of error, and additionally avoid bundling development code in production (or production code in development), we can use a minimal and conservative form of dead code removal, such as that provided by [babel-plugin-minify-dead-code-elimination](https://www.npmjs.com/package/babel-plugin-minify-dead-code-elimination).

In order to minimize the risk of unanticipated side-effects, we use the `keepFnName`, `keepFnArgs`, and `keepClassName` options, which ensures that we won't break code that depends on `fn.length`, `fn.name`, or `cls.name`.

Before using this plugin, we produce a file like:

```javascript
Liferay.Loader.define('frontend-js-react-web$react@16.8.6/index', ['module', 'exports', 'require', './cjs/react.production.min', './cjs/react.development'], function (module, exports, require) {
  var define = undefined;
  if (true) {
    module.exports = require('./cjs/react.production.min');
  } else {
    module.exports = require('./cjs/react.development');
  }
});
```

With it, we get:

```javascript
Liferay.Loader.define('frontend-js-react-web$react@16.8.6/index', ['module', 'exports', 'require', './cjs/react.production.min'], function (module, exports, require) {
  var define = undefined;

  module.exports = require('./cjs/react.production.min');
});
```

In other words, we strip out the unreachable branch, and additionally drop the unwanted "./cjs/react.development" from the list of dependencies.

Closes: https://github.com/liferay/liferay-js-toolkit/issues/379

Sample screenshot of running in liferay-portal in "development" mode:

New_User_Segment_-_Liferay

### DIFF
--- a/packages/babel-preset-liferay-standard/README.md
+++ b/packages/babel-preset-liferay-standard/README.md
@@ -24,7 +24,8 @@ This preset includes the following plugins:
 
 1. [babel-plugin-normalize-requires](https://github.com/izaera/liferay-js-toolkit/tree/master/packages/babel-plugin-normalize-requires)
 2. [babel-plugin-transform-node-env-inline](https://www.npmjs.com/package/babel-plugin-transform-node-env-inline)
-3. [babel-plugin-wrap-modules-amd](https://github.com/izaera/liferay-js-toolkit/tree/master/packages/babel-plugin-wrap-modules-amd)
-4. [babel-plugin-name-amd-modules](https://github.com/izaera/liferay-js-toolkit/tree/master/packages/babel-plugin-name-amd-modules)
-5. [babel-plugin-namespace-modules](https://github.com/izaera/liferay-js-toolkit/tree/master/packages/babel-plugin-namespace-modules)
-6. [babel-plugin-namespace-amd-define](https://github.com/izaera/liferay-js-toolkit/tree/master/packages/babel-plugin-namespace-amd-define)
+3. [babel-plugin-minify-dead-code-elimination](https://www.npmjs.com/package/babel-plugin-minify-dead-code-elimination)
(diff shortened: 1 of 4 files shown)


jbalsas:
Nice @wincent!

I think it's about time we moved this to yarn, don't you think @izaera? 😂

wincent:
> I think it's about time we moved this to yarn, don't you think @izaera? 😂

I wonder if the package-lock.json churn is because I am using `npm` v6.9.0, or because of unusually high solar flare activity today.

izaera:
Just started reviewing :)

PR #386, opened 2019-08-21
FIX #383

### LINKED ISSUE #383: EEXIST error in Windows when creating build.lr\liferay folder
I initially had issues converting an existing project, so I started from scratch:

1. create-react-app my-liferay-demo
2. yo liferay-js:adapt
3. npm run build:liferay

Below is the console output/error that I received. What other information is needed to help diagnose/troubleshoot this? I see content in /build and /build.liferay - but my assumption was that I should not see the error below? I also tried adding configuration to .npmbundlerrc to generate a jar at the end of the build, and it appears that it never gets that far (I believe confirming for me that something along the way is indeed erroring, as opposed to just some general notes at the end of an otherwise successful process).

```
> my-liferay-demo@0.1.0 build:liferay C:\dev\prototyping\my-liferay-demo
> lnbs-build

ℹ️ Making a backup of 'src/index.js' to 'src/.index.js' before
injecting Liferay JS Toolkit's modifications for the build.

yarn run v1.17.3
(node:4308) ExperimentalWarning: The fs.promises API is experimental
$ react-scripts build
(node:15244) ExperimentalWarning: The fs.promises API is experimental
Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  38.54 KB  build\static\js\2.7d980bde.chunk.js
  783 B     build\static\js\runtime~main.c7442541.js
  643 B     build\static\js\main.3b3d2171.chunk.js
  517 B     build\static\css\main.2cce8147.chunk.css

The project was built assuming it is hosted at the server root.
You can control this with the homepage field in your package.json.
For example, add this to build it for GitHub Pages:

  "homepage" : "http://myname.github.io/myapp",

The build folder is ready to be deployed.
You may serve it with a static server:

  npm install -g serve
  serve -s build

Find out more about deployment here:

  https://bit.ly/CRA-deploy

Done in 37.60s.
ℹ️ Restoring backup of 'src/index.js' after React's build has finished.

{ Error: EEXIST: file already exists, mkdir 'build.liferay\jar'
    at Object.fs.mkdirSync (fs.js:854:3)
    at Renderer.render (C:\dev\prototyping\my-liferay-demo\node_modules\liferay-npm-build-support\lib\util.js:85:17)
    at generateIndexJs (C:\dev\prototyping\my-liferay-demo\node_modules\liferay-npm-build-support\lib\scripts\build\create-react-app.js:257:11)
  errno: -4075,
  syscall: 'mkdir',
  code: 'EEXIST',
  path: 'build.liferay\\jar' }
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! my-liferay-demo@0.1.0 build:liferay: `lnbs-build`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the my-liferay-demo@0.1.0 build:liferay script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\e404rlr\AppData\Roaming\npm-cache\_logs\2019-08-19T20_22_00_122Z-debug.log
```

reinrl:
This change wraps the fs.mkdirSync() call in a try catch, explicitly swallowing error code EEXIST (the directory already exists, which seemed to be an issue on Windows) while bubbling up any other errors.

### DIFF
--- a/packages/liferay-npm-build-support/src/util.js
+++ b/packages/liferay-npm-build-support/src/util.js
@@ -7,7 +7,7 @@
 import child_process from 'child_process';
 import spawn from 'cross-spawn';
 import ejs from 'ejs';
-import fs from 'fs';
+import fs from 'fs-extra';
 import project from 'liferay-npm-build-tools-common/lib/project';
 import path from 'path';
 import resolveModule from 'resolve';
@@ -34,13 +34,7 @@ export class Renderer {
 
 		const outputPath = path.join(dir, name);
 
-		try {
-			fs.mkdirSync(path.dirname(outputPath), {recursive: true});
-		} catch (err) {
-			if (err.code !== 'EEXIST') {
-				throw err;
-			}
-		}
+		fs.ensureDirSync(path.dirname(outputPath));
 
 		ejs.renderFile(
 			path.join(this._templatesDir, `${template}.ejs`),
--- a/packages/liferay-npm-build-support/src/util.js
+++ b/packages/liferay-npm-build-support/src/util.js
@@ -34,7 +34,11 @@ export class Renderer {
 
 		const outputPath = path.join(dir, name);
 
-		fs.mkdirSync(path.dirname(outputPath), {recursive: true});
+		try {
+			fs.mkdirSync(path.dirname(outputPath), {recursive: true});
+		} catch (err) {
+			if (err.code !== 'EEXIST') throw err	
+		}
 
 		ejs.renderFile(
 			path.join(this._templatesDir, `${template}.ejs`),
--- a/packages/liferay-npm-build-support/src/util.js
+++ b/packages/liferay-npm-build-support/src/util.js
@@ -37,7 +37,9 @@ export class Renderer {
 		try {
 			fs.mkdirSync(path.dirname(outputPath), {recursive: true});
 		} catch (err) {
-			if (err.code !== 'EEXIST') throw err	
+			if (err.code !== 'EEXIST') {
+				throw err;
+			}
 		}
 
 		ejs.renderFile(


packages/liferay-npm-build-support/src/util.js
```diff
@@ -34,7 +34,11 @@ export class Renderer {
 
 		const outputPath = path.join(dir, name);
 
-		fs.mkdirSync(path.dirname(outputPath), {recursive: true});
+		try {
+			fs.mkdirSync(path.dirname(outputPath), {recursive: true});
+		} catch (err) {
+			if (err.code !== 'EEXIST') throw err
```
wincent: Devil's advocate: not all `EEXIST` errors can be assumed to be benign. For example, if I touch a file (not a directory) at the `outputPath`, then wouldn't we get (and swallow) `EEXIST` then too?

(Independently, a couple of style nits: want a trailing semi-colon on the statement, and I believe the pattern in this repo is to use braces around all blocks.)

reinrl: I corrected the style pieces, but your initial comment opens up what I believe is a broader discussion. From a bit of digging, I believe that a switch to something like [mkdirp](https://github.com/substack/node-mkdirp) would help alleviate that concern - but comes at the cost of a new dependency. Alternately, this piece of code is going after something that should only be generated by this process (or a previous run of it) - so is it an acceptable compromise (specifically to this case anyway) to recognize that a file named jar inside of a directory named build.liferay would return what is essentially a false positive? I believe that if a user manually created a file named jar (no extension) within /build.liferay, this would still error later in the process, as attempting to write files to a location that is actually itself a file would not work as intended.

wincent: Any decision like this  is essentially weighing two obviously desirable but mutually exclusive things against each other:

- "Fail fast": provide feedback as early as possible in the cycle.
- "Stay small": avoid taking on dependencies.

@izaera will have a better sense than I about where the balance lays in this project, but over time I find myself more and more reluctant to take on new dependencies unless they strongly prove their value. It's not just the way that a dependency can bring in a potentially large number of transitive dependencies, or the way that in recent years there have been several well-publicized incidents in which a popular dependency (often a deeply transitive one) had malicious code inserted. It's also the time it takes to vet/audit a dependency (and not just when you first add it, but whenever you update). Tools like `yarn audit`/`npm audit` help but they also make the costs of maintaining a large dependency footprint visible: in the old days a bad dependency might have gone unnoticed; now, these tools can lead to "drop-everything-and-uprade-now" scenarios — sometimes because the issue is real, and sometimes merely because of the optics, where a problem is actually non-critical but people disagree about whether that's the case because doesn't (and can't) provide a clear indication of whether an issue actually matters in a given context — when we'd rather be working on new features.

Obviously, one needs to strike a compromise — you don't want to spend all your time reinventing wheels for already solved problems — and there a plenty of projects that deliver enough value to be used, despite the downsides. My personal feeling in this concrete instance is that delaying the feedback from commit time to PR time isn't ideal, but it's still probably better than taking on new dependencies. I realize though that might be a bit contrarian considering how widely used Husky and lint-staged are.

izaera: @reinrl @wincent We had lint-staged a long time ago in place but removed it because it caused issues with some git workflows. Specifically, if you want to stage partial modifications in a file, lint-staged ends up staging the whole file and destroying all your partial staging work (which, BTW, can add up to a lot of time under some circumstances) :grimacing: 

In addition to that, we have the linting target which may be run locally and it's executed in the CI for every pull, so it shouldn't be so bad. I agree that you get feedback later in the CI as opposed to the build, but it's not so horrible. I agree with @wincent there too: making the build more complex and adding dependencies just to see an error 5 minutes before doesn't seem so valuable to me.

izaera: Regarding the `touch jar` thing I agree that it's a very unusual case that we may not cover. Why would anyone do that apart from trying to make the tool fail? :thinking: 

Also, if we want to cover it, we need to add much extra code and it's not clear what to do: should we delete the jar file and create a dir instead or just warn the user?

To finish with, the `build.lr` dir is considered our "exclusive property" (of the bundler I mean) so I would say we can always tell people not to use it for "their things" :sweat_smile: .

izaera:
BTW, I will merge this once #303 is in place. Sorry about the delay.

izaera:
Just started reviewing :)

izaera:
Resent as https://github.com/liferay/liferay-js-toolkit/pull/401
Will merge there.

Thx a lot @reinrl :-)

PR #395, opened 2019-09-05
Fixes #393: Create css-loader

### LINKED ISSUE #393: Create a css-loader to load CSS from remote URLs
We have `style-loader` to transform .css files into .js modules that insert the CSS directly into the HTML by means of a <style> tag.

Now we want the same but inserting a <link rel="stylesheet"> tag.

izaera:


packages/liferay-npm-bundler-loader-css-loader/src/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+	const href = getHref(filePath, extension, pathModule);
+
+	// Note that Liferay.ThemeDisplay.getPathContext() when called at runtime
+	// returns both pathProxy and the context path of the portal's webapp.
+	context.extraArtifacts[`${filePath}.js.wrap-modules-amd.template`] = `
+var link = document.createElement("link");
+link.setAttribute("rel", "stylesheet");
+link.setAttribute("type", "text/css");
+link.setAttribute("href", Liferay.ThemeDisplay.getPathContext() + "${href}");
```
jbalsas: I know we checked and `languageId` was getting retrieved from session, but are we sure we don't want it here?

izaera: Well, the rest of CSSs in the portal are downloaded without the `languageId` parameter so I assumed we should do the same... :man_shrugging:

jbalsas: 👍

jbalsas:
Since this is a bit overwhelming anyways to provide more useful feedback I'd say we've gone through enough iterations of this already so we can try it out in the field.

PR #398, opened 2019-09-05
Fixes #397: Prettier scripts do not run as is on Windows

### LINKED ISSUE #397: Prettier scripts do not run as is on Windows
From the main package.json, the two Prettier related scripts (format and format:check) result in an error like the following:

```
[error] No matching files. Patterns tried: '.*.js' 'packages/*/src/**/*.js' !**/node_modules/** !./node_modules/** !**/.{git,svn,hg}/** !./.{git,svn,hg}/**
```

This is likely due to how Windows inteprets quotes in NPM scripts. I will submit a pull request shortly...

reinrl:
Updated single quotes to escaped double quotes in format, format:check script definitions. This corrects the issue for Windows, can someone else confirm that it continues to function as expected on non-Windows machines?

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,8 @@
 	"scripts": {
 		"build": "lerna run build",
 		"ci": "npm run format:check && npm run lint && npm run test",
-		"format": "prettier --write '.*.js' 'packages/*/src/**/*.js'",
-		"format:check": "prettier --list-different '.*.js' 'packages/*/src/**/*.js'",
+		"format": "prettier --write \".*.js\" \"packages/*/src/**/*.js\"",
+		"format:check": "prettier --list-different \".*.js\" \"packages/*/src/**/*.js\"",
 		"lerna": "npm install && lerna bootstrap",
 		"lint": "eslint 'packages/*/src/*.js' 'packages/*/src/**/*.js'",
 		"lint:fix": "eslint --fix 'packages/*/src/*.js' 'packages/*/src/**/*.js'",


wincent:
Haven't tested this but I believe it should be ok. The purpose of the quotes is to prevent the globs from being expanded by the shell before handling them off to Prettier. You can see here that on a typical "non-Windows" shell (eg. Bash or Zsh) either kind of quote prevents expansion:

liferay-js-toolkit___ls

No idea why Windows ends up working with doubles but not singles; it looks like the single-quoted version is sending the quotes as part of the arguments (ie. Prettier sees `'.*.js'` instead of `.*.js`). This is probably an even clearer demonstration:

liferay-js-toolkit___echo

izaera:
:thinking: We had this error in some other place too, but I can't remember the final solution. The problem, as @wincent is that each shell interprets things the way it likes, and  as far as I remember, the other time we found this there were problems with alternative syntax.

I will check and come with more info :+1: .

izaera:
Just started reviewing :)

izaera:
Looks like I added single quotes in #262 which don't fix the error totally for Windows as we have seen.

So, I've tested this on linux and works fine. I assume the same result for Mac given that it uses bash too (in the end this is a shell script issue more than an OS one).

izaera:
Resent as https://github.com/liferay/liferay-js-toolkit/pull/402

Thx for the contribution @reinrl

PR #424, opened 2019-09-26
chore: update paths to reference new build locations

wincent:
As per: https://github.com/liferay/liferay-npm-tools/issues/272


wincent:
Closing this because we don't want to break client configs. We'll configure all of our modules to explicitly use the new path.

izaera:
Unfortunately, we cannot merge this change because those defaults are already a contract and we may break projects out there :man_shrugging: .

PR #426, opened 2019-09-30
Use --skip-install in QA scripts

izaera:


izaera:
I'm merging this as it is green, it's blocking me, and we are not releasing a new version yet, so we have a chance to modify anything we don't like in a further PR.

PR #428, opened 2019-09-30
Fixes #389: Check require calls scope

### LINKED ISSUE #389: The scope of `require` calls is not taken into account
Some packages are bundled with webpack in a format that has calls to the internal webpack bundle's `require` function. For example, `pako@1.0.10` comes with a `dist/pako_deflate.js` file that has `require` calls like:

```javascript
},{}],2:[function(require,module,exports){
// String encode/decode helpers
'use strict';


var utils = require('./common');
```

That call is clearly an internal reference to the webpack bundle that is translated into submodule `1` inside the webpack bundle:

```javascript
},{"./common":1}],3:[function(require,module,exports){
```

However, the bundler thinks that it is a legitimate `require` call and puts `./common` in the list of AMD dependencies of the module thus leading to an unresolved dependency error when the AMD loaders kicks in.

There are ways in babel to know if a `require` call is globally scoped and we should probably use that to ignore non-global `require` calls.

izaera:


izaera:
I'll go ahead and merge this since it has been tested and I don't see any side effects that could break anything.

PR #440, opened 2019-10-18
Fix dev server for liferay-npm-build-support

spiderpoul:
Since version 2.13.2 webpack dev server not working:
```
internal/validators.js:125
    throw new ERR_INVALID_ARG_TYPE(name, 'string', value);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type object
```

This fixes this error.


izaera:
@spiderpoul Thanks a lot for the contribution :-).

This was already fixed [in this merged PR](https://github.com/liferay/liferay-js-toolkit/pull/427/files). Sadly, we haven't released it yet because we are having complex issues with what appeared to be an innocent bug and it is delaying us :-(.

PR #484, opened 2019-12-30
Fix QA script

izaera:


scripts/qa/index.js
```diff
@@ -58,12 +58,12 @@ if (argv['deploy']) {
 	logStep('Deploying adapted project samples');
 	spawn('node', [lernaPath, 'run', 'deploy:liferay'], {
 		cwd: samplesDir,
-		env: {
+		env: Object.assign({}, process.env, {
```
jbalsas: Aren't we using `eslint-config-liferay` here? Because I think we wanted to prefer spread -> https://github.com/liferay/eslint-config-liferay/commit/b504cef328ea1cf38f37524588b5aab249fe8aca

jbalsas:
LGTM, @izaera, feel free to merge!

I think it would be good to leave a note somewhere as to why it needs to be done at `runtime` and not build-time, just to make sure we won't try to "optimize" it in the future ;)

scripts/qa/index.js
izaera: I tried the spread operator first but gave a syntax error and I wasn't able to fix it :man_shrugging: . Let me have a look again.

izaera: Yes. I had an old version of eslint-config-liferay :+1:

izaera:
That's explained here -> https://github.com/liferay/liferay-js-toolkit/pull/484/commits/b4cec3ae6aa48d0962ba62b615a99772e7ce73e5

PR #503, opened 2020-02-11
Prepare v2.18

izaera:


wincent:
Is the title right here? I thought 2.17.1 [landed in master 18 days ago](https://github.com/liferay/liferay-js-toolkit/commit/33e9c0c40f47bf8eb1b8b7ef497b92b36bf511e3).

izaera:
@wincent It is the automatic title someone generates (don't know who, though...). 

The PRs in this project are not meant to be semantic, because it only believes in issues and sees PRs as ephemeral artifacts, but it's OK to change their titles to make them more understandable. I'll do it.

The reason why the PR approach doesn't work too well is that several issues have been fixed in more than one PR so the safest way to track the decisions made is the issues and the git log. Also, the "changelog" is composed by the issues alone, and doesn't take PRs into account.

izaera:
OK, so the automatic title comes from the last commit in `master`which, by definition (because we have develop and master branches), is the previous released version :sweat_smile:

wincent:
> OK, so the automatic title comes from the last commit in `master`which, by definition (because we have develop and master branches), is the previous released version 😅

Makes perfect sense. 😀

PR #505, opened 2020-02-12
Fixes #504: Create correct CSS URL when using yarn

### LINKED ISSUE #504: CSS loader creates incorrect URLs when dependencies are hoisted by yarn
In a yarn setup it may happen that some dependency files live outside of the project folder, several levels up in the FS hierarchy.

In such cases, the CSS loaders fails to recognise the CSS file as an external dependency and composes a wrong URL like, for example:

```
/o/dynamic-data-mapping-form-field-type/../../../node_modules/leaflet/dist/leaflet.css
```

which is obviously wrong, as it should be more along the lines of:

```
/o/dynamic-data-mapping-form-field-type/node_modules/dynamic-data-mapping-form-field-type$leaflet@1.2.0/dist/leaflet.css
```

izaera:


izaera:
I'm merging this as it is blocking some developing in liferay-portal and seems low risk.

Issue #514, opened 2020-02-24
portletInstance Configuration label doesn't get translated if doesn't exists on language files 


jgarcialr:
When you create a new Liferay Vue application using liferay-js-toolkit that supports i10n, a new component is created.

If you create one field inside configuration at portletInstance level, and doesn't put that key inside Language_*.properties files, this label isn't shown in configuration UI of the module, instead there's a blank.

Per example: 

Given this config:
`"portletInstance": {
		"fields": {
			"drink": {
				"type": "string",
				"name": "Nambe-pinstance",
				"description": "drink-help",
				"default": "water",
				"options": {
					"water": "water",
					"wine": "wine",
					"beer": "beer"
				}
			}
		}
	}
`

If we don't put "Nambe-pinstance" inside Language.properties file, when we transform the config into portlet_preferences.json the result will be:

`{
  "availableLanguageIds": [],
  "fields": [
    {
      "name": "drink",
      "label": {
      },
      "dataType": "string",
      "type": "select",
      "tip": {
        "": "Choose the drink you like the most"
      },
      "predefinedValue": {
        "": "[\"water\"]"
      },
      "options": [
        {
          "value": "water",
          "label": {
            "": "Water"
          }
        },
        {
          "value": "wine",
          "label": {
            "": "Wine"
          }
        },
        {
          "value": "beer",
          "label": {
            "": "Beer"
          }
        }
      ]
    }
  ]
}`

And when you enter the UI, the label for the field with key "Nambe-pinstance" will be blank.

If we follow the same pattern as other modules, if the key wasn't exists in Language.properties files, the UI should show the original key instead of a blank.

PR #515, opened 2020-02-24
#514 fallback to field-key instead of blank when it doesn't exists on Language files

jgarcialr:
//cc @antonio-ortega


izaera:
Just started reviewing :)

izaera:
Resent as #529

PR #519, opened 2020-02-27
feat: merge code produced by `explicit exports PoC` 

izaera:


wincent:
I would have expected CI to complain about the license headers in some of these files because they don't match [the copyright.js template](https://github.com/liferay/liferay-js-toolkit/blob/3.x-WIP/copyright.js), but [it didn't](https://travis-ci.org/liferay/liferay-js-toolkit/builds/655790941). Maybe because it is running `eslint "packages/*/src/*.js" "packages/*/src/**/*.js"`, which means that it isn't checking any of the `.ts` files.

Getting ESLint to work properly with TypeScript requires a bit of tweaking; you could look at the config in [the Clay repo for an example](https://github.com/liferay/clay/blob/876c6b774dcb7e3e9d2cf02120b876a85abd46d2/.eslintrc.js#L15-L16) if you wanted.

izaera:
I still need to fix the build, but I pushed some more commits to backup my local modifications.

izaera:
All fixed but still not linting the copyright headers. Having a look at that.

Issue #520, opened 2020-02-27
Implement imports


izaera:
Now that we have explicits exports (#516) we need to be able to import packages from a different OSGI bundle.

The "problem" is that there's something incorrect with the imports in bundler 2.x that we can fix for bundler 3.x but at the cost of introducing a big interoperability burden that will be with us until everybody migrates everything to bundler 3.x. 

Let's see the problem. Imagine you create the `react-portlet` and `react-provider` projects (both at version `1.0.0`) where the second exports the React framework (version `16.12.0`) so that the first one imports it. 

This would lead to the following module defined in the AMD loader:

```
react-provider$react@16.12.0/index
```

Which would be imported by `react-portlet` by means of the `.npmbundlerrc`.

This looks good, but pay attention to the name of the module: it contains React's version (`16.12.0`) but not `react-provider`'s version (`1.0.0`). Thus, if we deploy another version of `react-provider` (which is legal in OSGi) there will be a collision if it also contains `react@16.12.0`.

We have never seen any problem related to this because the scenario is quite strange and even if it happened, the loader would choose one of the two `react@16.12.0` copies and, if both are the same (which is expected), nothing would happen.

However it is tempting to fix it for bundler 3.x, not only to avoid the error, but to simplify how the project is configured and built a lot (see next comment).

izaera:
In bundler 2, for imports to happen configuration is quite complex. In particular, you need to specify `react`'s version two times in `react-portlet`: one in `package.json` and another in `.npmbundlerrc`, like this:

**package.json**
```json
{
  "dependencies": {
    "react": "^16.0.0"
  }
}
```
**.npmbundlerrc**
```json
{
  "config": {
    "imports": {
      "react-provider": {
        "react": "^16.0.0"
      }
    }
  }
}
```

Note that in this case we use the same semver expression in both files, but it could be different because the first one is just used by the project during build time and the second one during runtime. You can even remove the dependency from the `package.json` file and nothing will fail during runtime!

This leads to several strategies to handle project and imports versions which adds complexity and separates from the standard expected frontend development practices (though it is obvious that we will need some deviation as the imports are not a standard feature). 

In an ideal world we would like to specify everything in `package.json` and simply tell the bundler where to get dependencies from during runtime. And then leave everything else to the bundler/platform.

We'll see how this can be achieved in bundler 3.x in the next comment.

izaera:
Imagine if we could be able to configure imports like this:

**package.json**
```json
{
[...]

PR #533, opened 2020-03-04
feat: #520 Implement imports

izaera:


packages/liferay-npm-build-tools-common/src/project/jar.ts
```diff
@@ -152,7 +152,7 @@ export default class Jar {
 		return prop.get(
 			npmbundlerrc,
 			'create-jar.features.js-extender',
-			pkgJson['portlet'] ? true : false
+			pkgJson.portlet ? true : false
```
wincent: Not a huge fan of ternaries of the form `? true : false`

Would you like `!!pkgJson.portlet` better?

packages/liferay-npm-bundler/src/steps/webpack/adapt/index.ts
```diff
@@ -0,0 +1,111 @@
...
+	await wrapBundles();
+	await injectImportsInPkgJson();
+}
+
+/**
+ * Copy bundles generated by webpack from webpack's output dir to bundler's.
+ */
+async function copyWebpackBundles(): Promise<void> {
+	await Promise.all(
+		['runtime', 'vendor', ...Object.keys(project.exports)].map(async id => {
```
wincent: We might want to run this in smaller batches because if the package exports a large number of items (I'm thinking of lodash as an example), we might wind up with a hundred concurrent tasks.

In any case, why is this `async` if you are calling blocking `Sync` functions in here?

packages/liferay-npm-bundler/src/log.ts
```diff
@@ -29,6 +31,14 @@ export function error(...args: unknown[]): void {
 	fmt.print(fmt.error`${args.join(' ')}`);
 }
 
+export function warn(...args: unknown[]): void {
+	if (!warnOn) {
+		return;
+	}
+
+	fmt.print(fmt.warn`${args.join(' ')}`);
```
wincent: (Over-the-shoulder comment from @jbalsas: why is this `fmt` and not `format`?) In general I think we eschew abbreviations other than very common ones (eg. `i` as a loop counter, or `pkg` because `package` is reserved etc).

.eslintignore
```diff
@@ -1,5 +1,7 @@
 /copyright.js
 node_modules/*
+packages/**/*.js
```
wincent: 😱 How many `.js` do we still have around?

packages/liferay-npm-build-tools-common/src/project/jar.ts
izaera: TBH I never decide which one I prefer :sweat_smile: Some days I prefer ?:, others !!

packages/liferay-npm-bundler/src/steps/webpack/adapt/index.ts
izaera: :thinking: It's probably a leftover of a refactor. I'd better change it to avoid the dreaded EMFILE error :sweat_smile: 

Off-topic: Regarding sync vs async file operations, I made some tests in bundler 2 when fixing the EMFILE error and noticed that sync was faster than async (even though it is counterintuitive). 

I suppose it has to do with OS caches and buffers, that make sync operations instantaneous and because you remove the async tasks handling overload from the code you end up with a faster version :man_shrugging: .

.eslintignore
izaera: :-D There are some projects not yet changed to the bundler 3 model. They will all disappear before the first release.

packages/liferay-npm-bundler/src/log.ts
izaera: Don't know why I abbreviated it :man_shrugging: I will change it.

packages/liferay-npm-build-tools-common/src/project/jar.ts
izaera: Changed.

packages/liferay-npm-bundler/src/steps/webpack/adapt/index.ts
wincent: My unsubstantiated hypothesis is that you probably don't get any benefit out of I/O concurrency when the files are so small (as they typically are in the case of JS source code). Maybe it would be different if the I/O was high latency (ie. dealing with network requests, for example).

izaera: Yep. Also, we are using SSD hard disk. Maybe in rotational HDs it changes...

PR #542, opened 2020-03-10
feat: Prepare a new snapshot version to integrate in Portal's build

izaera:


packages/liferay-npm-bundler/src/report/logger.ts
```diff
@@ -0,0 +1,82 @@
+/**
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+import project from 'liferay-npm-build-tools-common/lib/project';
+import {LogLevel} from 'liferay-npm-build-tools-common/lib/project/misc';
+
+let debugOn: boolean, infoOn: boolean, errorOn: boolean, warnOn: boolean;
```
wincent: We're going to [add a lint rule soon](https://github.com/liferay/eslint-config-liferay/issues/152) to require one statement per declaration, so you may as well start getting used to it now.

packages/liferay-npm-bundler/src/report/logger.ts
```diff
@@ -0,0 +1,82 @@
...
+	case 'info':
+		infoOn = true;
+	case 'warn':
+		warnOn = true;
+	case 'error':
+	default:
+		errorOn = true;
+	case 'off':
+}
+/* eslint-enable no-fallthrough */
```
wincent: Instead of all these variables and this creative `switch` statement, you could map log levels to numbers (just using the syslog numbers as an example):

```js
const LOG_LEVELS = {
  ERROR: 3,
  WARNING: 4
  INFO: 6,
  DEBUG: 7
};
```

and then your checks would be:

```js
if (logLevel > LOG_LEVELS.DEBUG)
```

etc

packages/liferay-npm-bundler/src/steps/webpack/plugin/imports-loader.ts
```diff
@@ -7,14 +7,27 @@ import {parse} from 'acorn';
...
 
 export default function(content: string): string {
-	if (content.indexOf('require') == -1) {
+	const log = report.getWebpackLogger(
+		'imports-loader',
+		getPrjRelPath(this.request)
+	);
+
+	// Early fail for performance: look for require/import
+	if (content.indexOf('require') == -1 && content.indexOf('import') == -1) {
```
wincent: Please use `===`/`!==` consistently. There are places in this PR where you're using double-equals and others with triple-equals, but it is safer to just use triple-equals everywhere (I know you know the difference and would never make a mistake, but as an example for the less-expert developers around you, it's just safer to stick with triple).

packages/liferay-npm-bundler/src/steps/webpack/plugin/imports-loader.ts
```diff
@@ -26,71 +39,183 @@ export default function(content: string): string {
...
+				break;
+
+			default:
+				break;
+		}
+
+		this._modified = this._modified || modified;
+	}
+
+	private _enterCallExpression(node: ESTree.CallExpression): boolean {
```
wincent: Did you give any thought to using [private fields instead](https://github.com/microsoft/TypeScript/pull/30829)? I am aware of the huge discussion around trade-offs in that PR, but I expect that over time people will pretty much converge on using private fields for everything because they are ie ES, not just in TS.

packages/liferay-npm-bundler/src/steps/webpack/plugin/imports-loader.ts
```diff
@@ -26,71 +39,183 @@ export default function(content: string): string {
...
+					return `var ${specifier.local.name} = __REQUIRE__('${namespacedModuleName}');`;
+
+				// no default
 			}
+		});
+
+		if (parentNode.type !== 'Program') {
+			log.error(
+				`Found 'import' statements not at the top of the file: they will be ignored`
+			);
```
wincent: I'd expect that this is unreachable code, right? It's a syntax error for an import declaration to be anywhere other than the top level. (Dynamic imports will be call expressions, so I wouldn't think they could turn up here.)

scripts/clean.js
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+const fs = require('fs-extra');
+
+try {
+	const json = fs.readJSONSync('tsconfig.json');
+
+	if (json.compilerOptions && json.compilerOptions.outDir) {
+		fs.removeSync(json.compilerOptions.outDir);
```
wincent: Kind of scary that an error or something malicious in a config file might trigger a potentially large deletion. Is there any kind of sanity checking that would be worth doing here?

packages/liferay-npm-build-tools-common/src/project/index.ts
```diff
@@ -451,6 +451,31 @@ export class Project {
 		}
 	}
 
+	/**
+	 * Get directory where work files must be placed.
+	 *
+	 * @remarks
```
wincent: Is this even a valid JSDoc tag? I can't see it listed in [the top-level index](https://jsdoc.app/) in their docs.

packages/liferay-npm-bundler/src/report/logger.ts
izaera: I thought it was already active :sweat_smile: . This seems the only place where I have not used one line per decl.

packages/liferay-npm-bundler/src/report/logger.ts
izaera: I think we can use a TS enum better.

wincent: There is some value in sticking as close to ES as possible, though.

packages/liferay-npm-bundler/src/steps/webpack/plugin/imports-loader.ts
izaera: I haven't looked at true private fields yet. I knew they existed but didn't consider it because the code is coming from plain JS and hasn't been pure TS until recently. Even like that, it still has some dirty hacks (f.e casts to `any` to access private fields) because I didn't want to spend an exaggerated amount of time in moving from JS to TS.

So I've been using typescript's `private`, in addition to `_` prefixes (which we already had in JS) to manage visibility. 

Once everything settles we may start moving into that direction, which seems quite interesting.

packages/liferay-npm-bundler/src/steps/webpack/plugin/imports-loader.ts
izaera: Nope. It's perfectly reachable because we are the ones parsing the JS code so there exists a possibility that we are fed incorrect JS. Webpack would probably complain about it too (haven't tried) and babel does for sure.

I agree it is probably something that will never get executed because that would usually mean that some package has been published incorrectly in npmjs.com, but if we can handle that situation it's better to show it to the user, I guess.

It can also happen in project's source code (not in npm deps) if the user is targeting modern browser and doesn't want to transpile but leave the imports to be processed webpack.

scripts/clean.js
izaera: We can check if the outDir is inside the project's source tree. Other than that, we can only apply heuristics...

In any case, if we misconfigure that dir, tsc will write a bunch of files there. You won't lose anything but it can also lead to problems...

packages/liferay-npm-build-tools-common/src/project/index.ts
izaera: It is apparently a valid TSDoc tag :point_right: https://github.com/Microsoft/tsdoc

packages/liferay-npm-bundler/src/steps/webpack/plugin/imports-loader.ts
wincent: No, I meant I would expect the parser to throw an error before getting here ([demonstration](https://astexplorer.net/#/gist/845fd73dda9cabe1edc4da6a2b9fd6bc/2ee4727dbe747c4ff19d72d2995eed78952d6af8)). But now I see you've got `allowImportExportEverywhere: true` configured, which explains why this code could indeed get here. Does it make sense to parse so leniently? 🤷‍♀

scripts/clean.js
wincent: > We can check if the outDir is inside the project's source tree.

Sounds good to me.

packages/liferay-npm-bundler/src/steps/webpack/plugin/imports-loader.ts
izaera: I'm configuring the parser to be as lenient as possible so that we are not the ones reporting the syntax errors to the end user :roll_eyes: .

The detailed explanation is that we insert the `imports-loader` in the webpack chain just to handle `import ... from ...` statements and nothing else. The fact that I need to transpile them (only when they are true imports) is because we don't want webpack to process them in that case. 

In any other case we just want to let the code pass through as much as possible and let other parts of the chain fail if they consider it is needed.

In the case of a misplaced `import` I decided to log an error in the report file to ease debugging. I don't want to simply ignore the error because if that situation arises I won't rewrite the import, which may lead to very difficult to diagnose errors. So, just because I'm returning prematurely in that case, I tell the user about it.

In the end, we should think of bundling as a compilation process, where the tool does its best effort to handle and recover from errors and, when not possible, shows the user as much information as possible to help him fixing them.

wincent: > So, just because I'm returning prematurely in that case, I tell the user about it.

If you can't produce a working result, wouldn't throwing a hard error be better? I don't see why you would ever want to `log` an error but not `throw` it. It is too easy for users to ignore errors amid all the console spew.

izaera: We don't know if it is a working result, that's the dilemma.

Historically we have switched from early failing to not saying anything, to just logging the problems (in the project) and no strategy proved absolutely better than the others so now I tend to do the intermediate solution (logging) and adjust it when we see the real problems people have outside.

Nevertheless, it's not a console log, which we don't use any more (or at least if we use them, we assume they won't be read :sweat_smile: ) it's a message in the HTML report file which, from what I've seen, people know how to activate when their build doesn't work. And in cases where they don't know, we tell them to do it. 

So, hopefully, if we dump that here, and the code gets hit any time in the future, it will be easy to catch it (though not instantaneous). On the other hand, if we failed the build, we could get a lot of false positives that could impact us more.

wincent: > We don't know if it is a working result, that's the dilemma.

How can an `import` statement not at the top level be anything other than a syntax error and therefore broken? If the browser even parses such a file — not even talking about trying to execute a particular code path within it — it *will* throw a `SyntaxError`.

By which I mean, rather than not "knowing it is a working result", wouldn't we "know it is *not* a working result"?

izaera: OK, let me explain how the `imports-loader` works and why I don't want to put the responsibility of failing the build on it.

The first attempt to rewrite imports (in this project) was simply substituting `require` by `__REQUIRE__` in requires pointing to imported modules. This was based on the assumption that webpack's core (the last piece run to bundle things after loaders and resolvers are executed) received CommonJS code as input.

(The idea behind that substitution is that webpack's core doesn't understand the `__REQUIRE__` and thus, leaves it untouched. Then, the bundler, after webpack is run, transforms those calls again into AMD requires handled by our AMD loader)

Then came some packages which transpile their files (before publishing to npmjs) but leave them in ES6+ import format (as opposed to CommonJS). And it turned out that webpack knows how to bundle these files too. 

(Note that webpack won't leave those import statements in the final bundle, but will transform them into `_webpack_require` calls. So, the key point is that CommonJS `require` and ES6+ `import` are a temporary thing that exists only during the build, but don't appear in the final output. In other words, the browser won't see any `import` or `require`)

Because of those packages, I needed to intercept those `import` statements and transform them to `__REQUIRE__` too. Note that I cannot simply rewrite the imported module literal, I need to convert the `import` statement into some other thing that webpack doesn't process, as I did with `require`s. 

But I don't have any alternative other than transpiling the `import` statement to a `__REQUIRE__` call expression. This situation is not symmetric to the one before: where I transformed a call expression (`require`) into another call expression (`__REQUIRE__`).

Because of this difference, I need to implement things in a different (more complex way) but still I want to be `imports-loader` as simple as possible. Ideally it should be a little more than a smart regexp substitution (in fact, I don't use regexp in the `require` case just to make things safer, but I was tempted for speed reasons).

To sum up, I would like to find a simpler way to intercept `import` statements that wouldn't require transforming the code but because I can't, I transform it but don't want to do anything beyond that. 

I don't want to fail the build, or tell the user about syntax errors in a place where nobody expects. I leave it to webpack's core or the previous loaders (babel or tsc for example).

You could argue that in that case, I shouldn't log anything either, and you are probably right. So, what I have thought would be much better than a log or a failed build would be to cover any situation where an `import` may appear (whether it is legal or not) and be capable of transforming it into a `__REQUIRE__` without failing or logging anything. 

That way we would achieve total safeness and wouldn't have to bother about wrong code or strange setups because we know we always replace an import statement by a call expression in the source code no matter where it appears.

I will have a look at this possibility in the next PRs.

Another question that may arise is: why bothering with illegal import statements not at the top of the file?

Well, you never know. In the past I've seen quite creative things in npm packages and the possibility to support `imports` in the middle of the code may be quite tempting for some people, so I just want to make sure that if somebody designs some magic loader or webpack plugin to handle those, we don't fail because of that.
 
Keep in mind that the only reason why in-the-middle `imports` are not allowed is to be able to do an static analysis of dependencies, not because it wouldn't be possible. In fact, if babel or node converts `import` statements to requires, that's the proof that you would be able to place them in the middle of the file if you wanted to.

wincent: > Another question that may arise is: why bothering with illegal import statements not at the top of the file?

It's not just "another question", it is literally *the* entire point of this thread. 

> Keep in mind that the only reason why in-the-middle `imports` are not allowed is to be able to do an static analysis of dependencies, not because it wouldn't be possible. In fact, if babel or node converts `import` statements to requires, that's the proof that you would be able to place them in the middle of the file if you wanted to.

The thing is, though, [the spec says](https://www.ecma-international.org/ecma-262/10.0/index.html#sec-modules) that  `import` statements may only appear at the top-level of a module, which is why Node.js, and every browser, will throw a `SyntaxError` as soon as it sees one in a function body. It's why [Babel throws](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABADwBQEpEG8BQjEwC2ADnAE5SIDkUAFjGAOZUDcOAvkA&debug=false&forceAllTransforms=true&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cenv&prettier=false&targets=&version=7.8.7&externalPlugins=) (`'import' and 'export' may only appear at the top level (2:2)`) when it sees a non-top-level `import`. And it's why webpack will report `Module parse failed: 'import' and 'export' may only appear at the top level (2:2)` if it sees a non-top-level `import`.

You suggest it's possible that this *may cease to be an error in the future*, but I don't think there's any evidence that that is even remotely likely. (Not saying it's impossible, because obviously there's no law of physics here that would prevent it, just saying it is unlikely.) The fact that we have [dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) as a standard mechanism for using `import` (calls not statements) outside the top-level suggests that it is just incredibly unlikely that what you're suggesting may happen, actually will happen. Your entire argument is based on the remotest of hypothetical possibilities, but the evidence just doesn't support it.

izaera: I understand  your point. Still, I don't want to be the one that fails the build. I see value in webpack's core complaining about it better than a loader that the user has not configured and doesn't know why is there (we inject it under the hood). 

In the past this kind of errors (with liferay code in the stack trace and a non-standard message source/format) has confused some of our users: they tend to ask us via support or forums, instead of googling it, which causes quite a big loss of time (for them and for us).

If there's a misplaced import webpack parser will complain about it (for sure) after `imports-loader` (if babel or tsc hasn't done it before it), so we don't have to care about it.

In case webpack's parser fails to signal the error for any strange reason (I can only think of a bug) we have the log as a safety net. But, to be sincere, I don't expect to see that log in any report file in my life.

packages/liferay-npm-bundler/src/report/logger.ts
wincent: In case anyone stumbles across this in the future, here is a list of alternatives: https://2ality.com/2020/02/enum-alternatives-typescript.html

Issue #545, opened 2020-03-17
Generated portlet won't display values because of wrong data type in configuration


IstvanD:
Replace the configuration file contents `features/configuration.json` with the following:
```
​{
 "$schema": "https://raw.githubusercontent.com/liferay/liferay-js-toolkit/master/resources/schemas/configuration.schema.json",

 "portletInstance": {
  "fields": {
      "drink": {
        "type": "float",
        "name": "drink",
        "description": "drink-help",
        "default": "10.5"
      }
  }
 }
}
```
Deploy the portlet. Open it's configuration. It won't show any input fields for "drink"

The reason is that in the generated .jar, in `features/portlet_preferences.json`, this part of the code
```
    {
      "name": "drink",
      "label": {},
      "dataType": "double",
      "type": "ddm-decimal",
      "tip": {
        "": "Choose the drink you like the most"
      },
      "predefinedValue": {
        "": "10.5"
      }
    }
```
has "ddm-decimal" set as "type".
Liferay uses soy to render the field. The rendering template in `modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/src/main/resources/META-INF/resources/Numeric/NumericRegister.soy`
```
{namespace NumericRegister}

{deltemplate PageRenderer.RegisterFieldType variant="'numeric'"}
	{call Numeric.render data="all" /}
{/deltemplate}
```
will only take "numeric" fields as it's params.
The "type" in `features/portlet_preferences.json` has to be generated as "numeric" instead of "ddm-decimal".

[LPS-109702](https://issues.liferay.com/browse/LPS-109702) has been opened on this issue.

izaera:
"Numeric" doesn't work either.

izaera:
It does, but I also need to change it in integer fields.

IstvanD:
Hi @izaera 

Maybe it is case sensitive? Have you tried "numeric" (without capital N)?

izaera:
@IstvanD we posted at the same time :sweat_smile:. See https://github.com/liferay/liferay-js-toolkit/issues/545#issuecomment-603819864

PR #547, opened 2020-03-25
fix: #545 Generated portlet won't display values because of wrong data type in configuration

izaera:
Simply change the literals we were writing to the JAR file because they were incorrect DDM values.


izaera:
Merging this because it is needed by support.

Issue #535, opened 2020-03-05
Bundler fails with WebpackOptionsValidationError if no exports are defined


izaera:
```
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.entry should be an non-empty object.
   -> Multiple entry bundles are created. The key is the chunk name. The value can be a string or an array.
    at Object.webpack [as default] (/home/ivan/Liferay/CE/liferay-portal/modules/apps/frontend-js/frontend-js-recharts/node_modules/webpack/lib/webpack.js:31:9)
    at Promise (/home/ivan/Liferay/CE/liferay-portal/modules/apps/frontend-js/frontend-js-recharts/node_modules/liferay-npm-bundler/lib/steps/webpack/run.js:13:43)
    at new Promise (<anonymous>)
    at Object.run [as default] (/home/ivan/Liferay/CE/liferay-portal/modules/apps/frontend-js/frontend-js-recharts/node_modules/liferay-npm-bundler/lib/steps/webpack/run.js:12:12)
    at Object.runWebpack [as default] (/home/ivan/Liferay/CE/liferay-portal/modules/apps/frontend-js/frontend-js-recharts/node_modules/liferay-npm-bundler/lib/steps/webpack/index.js:30:38)
    at Object.default_1 [as default] (/home/ivan/Liferay/CE/liferay-portal/modules/apps/frontend-js/frontend-js-recharts/node_modules/liferay-npm-bundler/lib/index.js:53:32)
    at Object.<anonymous> (/home/ivan/Liferay/CE/liferay-portal/modules/apps/frontend-js/frontend-js-recharts/node_modules/liferay-npm-bundler/bin/liferay-npm-bundler.js:32:32)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
```

PR #551, opened 2020-03-27
fix: #535 fail build if no exports configured

izaera:
It doesn't make sense to have no exports in a build because it would be unusable (as if you didn't define any entry in webpack).

### DIFF
--- a/packages/liferay-npm-bundler/src/steps/webpack/configure.ts
+++ b/packages/liferay-npm-bundler/src/steps/webpack/configure.ts
@@ -11,6 +11,7 @@ import webpack from 'webpack';
 
 import {buildGeneratedDir, buildWebpackDir} from '../../dirs';
 import * as log from '../../log';
+import {abort} from '../util';
 
 export default function configure(): webpack.Configuration {
 	// Get user's config
@@ -42,6 +43,14 @@ export default function configure(): webpack.Configuration {
 		},
 		{}
 	);
+
+	if (Object.keys(webpackConfig.entry).length === 0) {
+		abort(
+			'Please configure at least one export in the project ' +
+				`(or add a 'main' entry to your package.json, or create an ` +
+				`'index.js' file in the project's folder)`
+		);
+	}
 
 	// Override output configuration
 	overrideWarn('output', webpackConfig.output);


packages/liferay-npm-bundler/src/steps/webpack/configure.ts
```diff
@@ -43,6 +44,14 @@ export default function configure(): webpack.Configuration {
 		{}
 	);
 
+	if (Object.keys(webpackConfig.entry).length === 0) {
```
wincent: Should this be?

```js
if (
  !webpackgConfig.entry ||
  typeof webpackConfig.entry !== 'object' ||
  !Object.keys(webpackConfig.entry).length
) {
```
to catch when `entry` is not supplied at all? (I can't see any guarantee via types that it will be present, but I am looking in GitHub, so I might be missing it.) Your error text below makes it look like omitting it may be fine, but in that case there must be some guarantee elsewhere that we won't wind up exploding here by calling `Object.keys(undefined)`. 🤷‍♂

izaera: It is guaranteed by the code one line above of the if: the `reduce`.

But given that TS won't force it, we may add those additional checks just in case. It will make code more difficult to read, but more robust...

wincent: This is what I meant about me not seeing it in GitHub... conveniently they hide the `reduce`, and when I searched the repo looking for related types, nothing turned up... 🤦‍♂️

fix___535_fail_build_if_no_exports_configured_by_izaera_·_Pull_Request__551_·_liferay_liferay-js-toolkit

izaera: So what? Do we add the extra conditions or leave it like it is? 

I prefer readability to robustness in this case. Given that it is one line above and any change to that code will affect both lines with high probability.

wincent: Whatever you want.

In an ideal world, my number 1 preference would be to have the type system verify for us that the webpack configuration has the expected shape. Do that once at the perimiter of the application where we first read the config, and from then on you can be sure that you always have a valid value of type `WebpackConfig`, so you don't need to scatter checks all over the codebase. But to do that properly, you need to make [an assertion function](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions), and that is probably more work than you want to do right now (especially because we may have to deal with imperfect configs and get into the messy business of deciding what is a hard error and what is something that could be coerced into a valid format).

So my number 2 preference would be a simple check *above* the `reduce` (because that's where, I think, you first assume the `entry` is present and a non-null object) . I don't think it harms readability too much to add an `if (!webpackConfig.entry || typeof webpackConfig.entry !== 'object')` check.

Preference number 3: do nothing.

izaera: The "problem" here is that we don't control the type, it is directly imported from `webpack` package. And in webpack that type has an optional `entry` :-(.

Regarding the "check above the reduce": I don't assume anywhere that `entry` is present. I derive it from `exports` (using `reduce`) which is always defined because it's always an object of type `Exports` :-).

izaera: I will merge as it is, then.

Issue #553, opened 2020-03-31
Modules are not sandboxed in an IIFE


izaera:
The code for JS modules is not wrapped in an [IIFE](http://adripofjavascript.com/blog/drips/an-introduction-to-iffes-immediately-invoked-function-expressions.html). 

This causes code outside of `Liferay.Loader.define()` (as the one generated by `css-loader`) to be executed in the global scope, making all function and variables declarations available in the global scope too.

The situation is specially bad when combo servlet gathers multiple module files and sends them in one URL. In such cases things like [LPS-111045](https://issues.liferay.com/browse/LPS-111045) may happen:

1) More than one CSS module is loaded
2) Because `css-loader` writes code like:

```javascript
var link = document.createElement("link");
link.setAttribute("rel", "stylesheet");
link.setAttribute("type", "text/css");
link.setAttribute("href", Liferay.ThemeDisplay.getPathContext() + "/o/fragment-web/node_modules/fragment-web$codemirror@5.52.0/lib/codemirror.css");

function defineModule() {
  Liferay.Loader.define("fragment-web$codemirror@5.52.0/lib/codemirror.css", ['module', 'exports', 'require'], function (module, exports, require) {
    var define = undefined;

    module.exports = link;
  });
}

link.onload = defineModule;
```

and `defineModule` is defined in the global scope, several CSS modules end up calling the same `defineModule`, which makes the loader complain.

izaera:
I will sandbox the define only when a wrap template is applied (to save bits).

PR #554, opened 2020-04-01
fix: #553 wrap module definitions into an IIFE

izaera:
See issue #553 for a full description of what happened and the commit for how it's fixed.

### DIFF
--- a/packages/babel-plugin-wrap-modules-amd/src/index.js
+++ b/packages/babel-plugin-wrap-modules-amd/src/index.js
@@ -135,9 +135,11 @@ function applyUserDefinedTemplateIfPresent(filename, defineNode, log) {
 
 	log.info('wrap-modules-amd', 'Applied user template to wrap file');
 
-	const buildUserTemplate = template(
-		fs.readFileSync(templateFile).toString()
-	);
+	const buildUserTemplate = template(`
+		(function() {
+			${fs.readFileSync(templateFile).toString()}
+		})();
+	`);
 
 	fs.unlinkSync(templateFile);
 


jbalsas:
LGTM

packages/babel-plugin-wrap-modules-amd/src/index.js
```diff
@@ -135,9 +135,11 @@ function applyUserDefinedTemplateIfPresent(filename, defineNode, log) {
 
 	log.info('wrap-modules-amd', 'Applied user template to wrap file');
 
-	const buildUserTemplate = template(
-		fs.readFileSync(templateFile).toString()
-	);
+	const buildUserTemplate = template(`
+		(function() {
+			${fs.readFileSync(templateFile).toString()}
```
jbalsas, line 140: Should we also try-catch'it?

izaera: :thinking: I would say no because:

1) If we catch the error and report it we are doing the same the browser does, so there's no point in doing it
2) If we catch the error and swallow it, the definition won't take place and it will fail the resolution with an obscure timeout error which is harder to debug.

jbalsas: The problem is, with combo, an error breaks the whole combo execution...

izaera: Yes, but still it will fail afterwards, when resolving the resolution. One combo request cannot belong to more than one `require()` resolution so, if the combo doesn't fail, the loader will, as I said.

It's what was happening in this error: the definition of two modules was done with the same name. That led to: 

1) an error in the console informing of a duplicate definition (because an exception is thrown inside the loader when a module is defined twice)
2) a following error stating a resolution timeout (the huge red message that everybody skips and forwards us :sweat_smile: )

If we had catched the double definition exception and swallowed it, it would have taken much more than one day to find the root cause. On the other hand, if we had catched and printed it, we would have been in exactly the same situation: as soon as one single module in a combo request fails to define, that `require()` is doomed.

But, if you feel it is more maintainable to catch and print to console.error (in case we reuse combo requests in the future) we can do it. However, it will impact module sizes a bit...

jbalsas: I never said `swallow` it 😉

jbalsas: But it's okay, let's push for now, since this has been the status quo for a long time

wincent: This is going to include some awkward whitespace now, isn't it?

izaera: @jbalsas I never said you said `swallow` it ;-P I just analyzed all the possibilities we had to handle the exception

@wincent Whitespace? If you mean the formatting of the IIFE, it doesn't show in the final output because it is converted into an AST prior to printing.

wincent: > Whitespace? If you mean the formatting of the IIFE

Yeah, that's exactly what I was referring to.

> it doesn't show in the final output because it is converted into an AST prior to printing.

Good to know. 👍

PR #566, opened 2020-05-07
chore: configure eslint for .js files

izaera:
I needed to reconfigure eslint to make it behave differently for .js and .ts files and, at the same time, integrate well with vscode so that it lints while developing. The result has been a big headache and this PR.

Hope everybody looking at it enjoys it as much as I've enjoyed developing it.


.eslintrc-js.js
```diff
@@ -0,0 +1,20 @@
...
+		'no-for-of-loops/no-for-of-loops': 'off',
+		'no-return-assign': ['error', 'except-parens'],
+		'notice/notice': [
+			'error',
+			{
+				templateFile: path.join(__dirname, 'copyright.js'),
+			},
+		],
+	},
+};
```
wincent, line 20: How is this file involved? I don't see any explicit references to it.

.eslintrc.js
```diff
@@ -35,4 +70,32 @@ module.exports = {
 			},
 		],
 	},
+
+	overrides: [
+		{
+			files: ['*.js'],
+			rules: {},
```
wincent, line 77: I would expect these two lines to have no effect. As in, even without this, I would expect the default "liferay" rules to apply to "*.js" files anyway. What's going on here?

.eslintrc-js.js
izaera: Oops... I have changed my usual pre-PR workflow recently and it seems that I should go back to what I did before because I'm committing too many unneeded things lately :-(.

This is a leftover of one of the multiple tests I had to do. I had two separate files: one for JS, another for TS. Then I found that I could make override work correctly and discarded the idea. But forgot to amend the commit :-(.

.eslintrc.js
izaera: We can remove it. It does nothing, but I wanted to leave it as a placeholder to remind us, in the future, that any specific JS rule goes there.

However, it is probably very unlikely that we will have a JS specific rule in the future, so I'll remove it if it causes confusion.

.eslintrc-js.js
wincent: > I have changed my usual pre-PR workflow recently

Just do what I do an use this alias:

```
alias go-go-go='sudo git add --force --chmod=+x --ignore-errors . && sudo git commit -m WIP --allow-empty && sudo git push origin +HEAD:master'
```

Issue #567, opened 2020-05-08
IE11 complains: module.exports is not a valid JSON in adapt-rt.js


izaera:
The problem is that the code we are shipping is ES6+ syntax and IE11 doesn't recognize that.

PR #568, opened 2020-05-08
fix: #567 IE11 complains: module.exports is not a valid JSON in adapt-rt.js

izaera:
See issue #567 for more context


izaera:
Merging this as I'm not the creator, but the reviewer.

Also, @jorgediaz-lr has reported that the customer having the issue has confirmed that the fix works.

Issue #526, opened 2020-02-27
Rename liferay-npm-build-tools-common to liferay-js-toolkit-core


izaera:
And place all common code there (even the one used by generators)

izaera:
As per [this commentary](https://github.com/liferay/liferay-js-toolkit/pull/563#discussion_r419976387) design a proper API for this package.

izaera:
@wincent What would be better as a public API for `liferay-js-toolkit-core?

Keep in mind that it will have different domains which are independent between them. For example:

1) one domain of utilities to read project configuration (the `Project` class)
2) one domain of utilities to transform files in functional fashion (JS, JSON, text, CSS, etc.)
3) TS types describing external file formats and/or protocols (with Liferay portal, for example)
4) logging utilities
5) millenial/hipster printing to the console
6) ...

Should the API be:

1) consumers always `import from 'liferay-js-toolkit-core'` and its main module exports a complex object that describes the domains and utilities and multiplexes into internal files
2) consumers import from different "logical" paths that describe the domains and utilities
3) a mix of 1 and 2 (limiting the depths of logical paths)
4) other

Examples would be:

1) 
```typescript
import {log, transform} from  'liferay-js-toolkit-core';

log.debug('hi');
transform.js.transformSource('var a', tranform.js.operation.wrapModule());
```

2) 
```typescript
import log from  'liferay-js-toolkit-core/log';
import {transformSource} from  'liferay-js-toolkit-core/transform/js';
import {wrapModule} from  'liferay-js-toolkit-core/transform/js/operation';

log.debug('hi');
transform.js.transformSource('var a', tranform.js.operation.wrapModule());
```

3) 
```typescript
import log from  'liferay-js-toolkit-core/log';
import transformJs from  'liferay-js-toolkit-core/transform/js';

log.debug('hi');
transformJs.transformSource('var a', transformJs.operation.wrapModule());
```

4)
:shrug:

izaera:
Note that when I talk about some domains inside the package I mean `some`, not `several` or `hundreds`.

This package should not be a box where every utility is dumped, but it will be, for sure, a package with more than one independent "subpackages". The common thing between them is that they will all be used for the same problem (JS projects inside the Liferay domain).

But we need one package to share code for the generators, tools, utilities, whatever we make (like JS themes generator, JS themes build, JS portlets build, ....) and we want just one package, not one per independent domain to ease releases.

wincent:
[...]

PR #570, opened 2020-05-11
feat: #526 rename liferay-npm-build-tools-common to liferay-js-toolkit-core

izaera:
This is the first part of the two PRs to implement #526. 

This PR is intended to get feedback on the export structure of `liferay-js-toolkit-core` which will be considered public API once we release bundler 3.

If we agree on this API structure, I will then refactor the code to put every common code out of `js-toolkit-core` (in `generator-liferay-js` for example) in it and clean as much as I can. Also, I will rename things in `liferay-js-toolkit-core` to avoid `as` statements in index.ts (we don't want to have aliases in the very first version).


packages/liferay-js-toolkit-core/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+export {babelIpc};
+export * from './babel-util'; // TODO: not sure if this is needed in bundler 3
+export {default as PkgDesc} from './pkg-desc';
+export {default as PluginLogger} from './plugin-logger';
+
+// Project descriptor class and types
+export * from './project';
+
+// Format library
+export {format};
```
wincent, line 36: This should be possible to write as:

```js
export * as format from './format';
```

Note that it may depend on Babel/TypeScript versions and settings. I [pointed this out over on Clay](https://github.com/liferay/clay/pull/3183#pullrequestreview-402608050) but Bryce had trouble getting it working. But it is definitely possible (I've [done it](https://github.com/wincent/wincent/blob/8f1a39222f03f9a751d081e2c9ad78e0e1f340a2/fig/index.ts#L15) in another project), and is described on [the MDN page for `export` statements](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export).

packages/liferay-js-toolkit-core/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+// Miscellaneous utilities
+export {negate as negateGlobs, prefix as prefixGlobs} from './globs';
+export * from './regexp';
```
wincent, line 40: All of these `*`, while a convenient shorthand, hide what's actually being exported. In the case of a module like `regexp.ts` (which I chose at random), I'd rather:

- Rename the file to `escapeStringRegexp.ts` because it only exports a single function and it is nice if the file name matches the contents.
- Make the export a `default` export.
- Change this to `export {default as escapeStringRegexp} from './escapeStringRegexp'` so that it's transparent.

packages/liferay-js-toolkit-core/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+// JSON file structure definitions (schemas)
+export * from './api/configuration-json';
+export * from './api/manifest';
```
wincent, line 44: This one is an example fo one that only exports types. TypeScript 3.8 makes it possible to use specific syntax for exporting and importing types because it solves a number of problems (see [the blog post](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8/)). It might be worth switching to that syntax (I think the explicitness of it is good).

packages/liferay-js-toolkit-core/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+// Handler for manifest.json files
+export {default as Manifest} from './manifest';
+
+// TODO: maybe remove the next section before bundler 3?
+// Utilities to deal with npm packages' information structures
+export * from './alias';
+export {normalizeImportsConfig, unrollImportsConfig} from './imports'; // TODO: remove imports for sure
+export * from './modules';
+export * from './namespace';
```
wincent, line 20: Taking this one as an example that exports multiple things, here's another reason why we might want to eschew the `*` shortcut and list the exports explicitly: it permits us to selectively make a subset of the module's functions "public", while the rest can be omitted and reserved for internal use. In general, we want to make public only what we have to, to maximize our freedom to refactor in the future.

packages/liferay-js-toolkit-core/src/index.ts
izaera: Yes. It's a tsc version issue because vscode understands it but our build doesn't. I will update tsc version and see what happens.

packages/liferay-js-toolkit-core/src/index.ts
izaera: Of course. In the final release I will export everything explicitly. It's more verbose, but it's the only way to see the API at a glance and also make sure that we don't export things we don't want.

There's the risk that we forget to export something we want but that would be fixed by releasing a new minor version. 

The opposite (exporting something we don't want by mistake) is much worse because someone could start using it and that would make more difficult (for us) to remove it.

packages/liferay-js-toolkit-core/src/index.ts
izaera: Yes.

Issue #562, opened 2020-04-30
Angular adapted app: Production environment is hardcoded in Liferay builder


ChrisMayor:
Currently, it is not possible for an Angular adapted app to have a build output from the Liferay builder, because always "ng build --prod" is called.

This is caused by the liferay-npm-build-support npm package, where the production build is hardcoded:

liferay-npm-build-support/src/scripts/build/index.js

```
case ProjectType.ANGULAR_CLI:
			buildWith('build', ['--prod=true']);
			break;
```

It would be very helpful for Liferay installations with multiple environments if the builder would allow different environment builds for adapted Angular apps.

izaera:
The only caveat here is that when not using the `--prod=true` switch, the output of `ng build` is totally different, so supporting it is akin to creating a new adaptation like those for React or Vue.

I mean, is not just a question of having a new configuration flag and an if in the code.

ChrisMayor:
Hello @izaera,
the output of the "prod=true" build is not necessarily different from another build configuration in angular. (ok in the default configuration of an angular app you are right ;-))

The parameter -prod=true is just a shorthand for "--configuration=production".

In our case, we needed different "production like builds" for the Angular portlet for our different environments. For example if you have an Angular Portlet with some kind of external backend REST API communication and 3 different environments (prod, QA, test) where you want to deploy your code. Then you would put the url to the REST Apis in the evironment config and add another environment (with specific config) to the angular.json file:

### Test

**angular.json:**
```
"configurations": {
  "test": {
    "sourceMap": {
      "scripts": true,
      "hidden": true
    }
  }
}
```

**environment.test,ts:**
```
{
apiUrl: https://test-api
}
```

### QA

**angular.json:**
```
"configurations": {
  "qa": {
    "sourceMap": {
      "scripts": true,
      "hidden": true
    }
  }
}
```

**environment.qa,ts:**
```
{
    apiUrl: https://qa-api
}
```

### Prod

**angular.json:**
```
"configurations": {
  "production": {
 ...
  }
}
```

**environment.prod,ts:**
```
{
    apiUrl: https://prod-api
}
```

In our case, we solved our problem by creating a custom build that wraps the Liferay build and copies the angular.json (with the environment as production) for the required environment before executing build:liferay.

wincent:
Going to do a mass-transfer of issues (including this one):

- **From:** https://github.com/liferay/liferay-js-toolkit/issues
[...]

PR #572, opened 2020-05-13
feat: #562 redesign liferay-js-toolkit-core API

izaera:
This is the second part of the core API polishing. We will refactor/move/remove more things later as we need it, but I don't want to do it now because many things will be deleted and there's no point in organizing this more if we are going to delete it after.

Pretty much everything in this commit is removal of code or internal reorganization inside the project usually not affecting the public API (in fact, this serves as a test of the index as manifest).

I have renamed some public API things (like schemas) to avoid naming collision and also make the API more uniform.


packages/liferay-js-toolkit-core/src/bundler/PkgDesc.ts
```diff
@@ -107,3 +107,5 @@ export default class PkgDesc {
 	private readonly _name: string;
 	private readonly _version: string;
 }
+
+export default PkgDesc;
```
wincent, line 111: This seems wrong (having a named export and then re-exporting it redundantly as the default export). I am guessing you wanted there to just be a default export.

packages/liferay-js-toolkit-core/src/bundler/PluginLogger.ts
```diff
@@ -156,3 +156,5 @@ export default class PluginLogger {
 
 	private _msgs: Message[];
 }
+
+export default PluginLogger;
```
wincent, line 160: Same thing here (default + named). There was some discussion about this being undesirable [over here](https://github.com/liferay/liferay-frontend-guidelines/issues/21).

packages/liferay-js-toolkit-core/src/project/Adapt.ts
```diff
@@ -58,3 +59,5 @@ export default class Adapt {
 	private readonly _project: Project;
 	private _supported: boolean;
 }
+
+export default Adapt;
```
wincent, line 63: Redundant named + default exports.

packages/liferay-js-toolkit-core/src/project/Jar.ts
```diff
@@ -231,3 +231,5 @@ the one in .npmbundlerrc
 	private readonly _project: Project;
 	private _webContextPath: string;
 }
+
+export default Jar;
```
wincent, line 235: Redundant named + default exports.

packages/liferay-js-toolkit-core/src/project/Localization.ts
```diff
@@ -163,3 +163,5 @@ export default class Localization {
 	private _languageFileBaseName: FilePath;
 	private _localizationFileMap: LocaleFilePathMap;
 }
+
+export default Localization;
```
wincent, line 167: Redundant named + default exports.

packages/liferay-js-toolkit-core/src/project/Misc.ts
```diff
@@ -154,3 +154,5 @@ export default class Misc {
 	private _reportFile: FilePath;
 	private _reportLevel: LogLevel;
 }
+
+export default Misc;
```
wincent, line 158: Redundant named + default exports.

packages/liferay-js-toolkit-core/src/project/Probe.ts
```diff
@@ -59,3 +59,5 @@ export default class Probe {
 
 	private readonly _project: Project;
 }
+
+export default Probe;
```
wincent, line 63: Redundant named + default exports.

packages/liferay-js-toolkit-core/src/project/Project.ts
```diff
@@ -549,3 +525,5 @@ export class Project {
 
 	private _workDir: FilePath;
 }
+
+export default Project;
```
wincent, line 529: Redundant named + default exports. (Yes, I know I'm repeating myself a lot — but we'll probably adding a lint against this, so you're setting yourself up for pain if you go this way.)

packages/liferay-js-toolkit-core/src/project/Rules.ts
```diff
@@ -256,3 +260,5 @@ export default class Rules {
 	private _rules: BundlerNormalizedRule[];
 	private _versionsInfo: Map<string, VersionInfo>;
 }
+
+export default Rules;
```
wincent, line 264: Redundant named + default exports.

packages/liferay-js-toolkit-core/src/project/VersionInfo.ts
```diff
@@ -0,0 +1,12 @@
...
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+/** Plugin version information */
+export interface VersionInfo {
+	version: string;
+	path: string;
+}
+
+export default VersionInfo;
```
wincent, line 12: Redundant named + default exports.

packages/liferay-js-toolkit-core/src/schema/ConfigurationJson.ts
```diff
@@ -27,3 +27,5 @@ export interface ConfigurationField {
 	required?: boolean;
 	type: 'string' | 'number' | 'float' | 'boolean' | 'password';
 }
+
+export default ConfigurationJson;
```
wincent, line 31: Redundant named + default exports.

packages/liferay-js-toolkit-core/src/schema/ManifestJson.ts
```diff
@@ -4,42 +4,44 @@
...
 }
 
 /** Structure of module flags in package descriptor in `manifest.json` files */
-export interface ModuleFlags {
+export interface ManifestJsonModuleFlags {
 	/** Module exports `__esModule` flag (as defined by Babel) */
 	esModule?: boolean;
 }
+
+export default ManifestJson;
```
wincent, line 47: Redundant named + default exports.

packages/liferay-js-toolkit-core/src/schema/PkgJson.ts
```diff
@@ -0,0 +1,26 @@
...
+	};
+	devDependencies?: {
+		[pkgName: string]: string;
+	};
+	peerDependencies?: {
+		[pkgName: string]: string;
+	};
+}
+
+export default PkgJson;
```
wincent, line 26: Redundant named + default exports.

izaera:
:thinking: Truth is I exported both as default and named to be able to
do:

```javascript
export * from './file';
```

in the manifest `index.js`.

We talked about exporting everything explicitly in the final release so
this double export won't be necessary any more.

However, I'm a bit hesitant about removing all `export *`s because in
some cases it seems safer to export everything inside a file (for
example, everything in `Project.ts` because the whole file is supposed
to be exported always; it doesn't make sense to have a type there if it
is not exported in the API). 

If we list everything explicitly in the manifest index, I'm pretty sure
that we will forget to export newly added types sometime and that will
lead to a patch release.

Note that even if we remove the default export, we may still `export
*`, we simply need to change every:

```javascript
import Project from '.../Project';
```

to 

```javascript
import {Project} from '.../Project';
```

It's just that to me it seemed a bit ugly as the file is named
`Project` but you still need to import it as named. But I can live with
it.

Splitting the `Project.ts` file into little pieces is very much
overkill in this case as the exported types are satellites of `Project`
with no real entity outside of it. 

I simply export them so that people may use them in their methods and
variables in case they want to pass any internal value of `Project`
itself.

One more thing that lead me to the double export is that it doesn't do
harm and remains internal to the project (it's not propagated to the
consumers). 

It is also a common practice in some node libraries, though I agree it
has more to do with CommonJS-ES6 interoperability than a real design
decission.


So, what do you think? Should we remove the default export and leave
just named ones?

wincent:
> If we list everything explicitly in the manifest index, I'm pretty sure that we will forget to export newly added types sometime and that will lead to a patch release.

Yeah, maybe. But that doesn't seem like a huge downside compared to the opposite scenario: ie. if you export something private by mistake, somebody starts using it, months later we break it and they complain to us, and we find out that for months they've been using a "private" API that we accidentally made public.

> ```javascript
> import Project from '.../Project';
> ```
>
> to
>
> ```javascript
> import {Project} from '.../Project';
> ```
>
> It's just that to me it seemed a bit ugly as the file is named `Project` but you still need to import it as named. But I can live with it.

Well, I don't think that's the case, is it? You can export `Project` as the `default` from `Project.ts` (don't export a named equivalent) and internally you can import it like:

```js
import Project from './Project';
```

and in the manifest you can re-export like this:

```js
export {default as Project} from './Project';
```

The manifest is the *one* place where you'd have to do this `default as` dance, and everywhere else it's easy (ie. internally, you just import the default, and in other packages, you just import the named export from the manifest).

> Splitting the `Project.ts` file into little pieces is very much overkill in this case

Yeah, I don't think I would do that.

> One more thing that lead me to the double export is that it doesn't do harm and remains internal to the project (it's not propagated to the consumers).

Well, it's clutter, unnecessary, and you're going to run afoul of the linter (when we add it). It simply doesn't make sense to export the same thing twice. IMO, exporting something as the default *and* as a named export is just as wrong as exporting the same thing under two names:

```
// This is pointless and wrong...
export default function foo() {}
export foo;

// Just like this is...
function fn() {}
export const a = fn;
export const b = fn;
```

> It is also a common practice in some node libraries, though I agree it has more to do with CommonJS-ES6 interoperability than a real design decision.

Which libraries? I'd never seen this anywhere until I came to Liferay and I was appalled by it.

> Should we remove the default export and leave just named ones?

I think we should do what [we say in the guidelines](https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/file_names.md):

> | File contents                                 | File name                   |
> | --------------------------------------------- | --------------------------- |
> | A class, `ThingParser`                        | `ThingParser.js`            |
> | A component, `GizmoPicker`                    | `GizmoPicker.js`            |
> | A constructor function, `Blob`                | `Blob.js`                   |
> | A function, `concatenateStrings`              | `concatenateStrings.js`     |
> | Multiple exports, one of which is the default | Based on the default export |
> | Multiple exports, with no default             | Descriptive CamelCase       |
> | Tests for file `src/**/*.js`                  | `test/**/*.js`              |

ie.

- If your module exports one thing, it should be the default export (only).
- If your module exports multiple things, they obviously have to be named exports, but:
  - If the is one overarching important thing that warrants being a default, it should be the default (only).
  - If not, there should only be named exports.

izaera:
> Well, I don't think that's the case, is it? You can export Project as
> the default from Project.ts (don't export a named equivalent) and
> internally you can import it like:
> 
> import Project from './Project';
> and in the manifest you can re-export like this:
> 
> export {default as Project} from './Project';
> The manifest is the one place where you'd have to do this default as
> dance, and everywhere else it's easy (ie. internally, you just import
> the default, and in other packages, you just import the named export
> from the manifest).
> 

🤔 Let's see... If we write this neither tsc nor eslint complain:

export * from './project/Project';
export {default as Project} from './project/Project';


So I guess we can live with that.

> Which libraries? I'd never seen this anywhere until I came to Liferay
> and I was appalled by it.

I can't think of any now. But rethinking this I may be referring to
other thing (similar but slightly different to this case): 

There are some libraries that export the same thing both as
module.exports.default and module.exports to ease interoperability
between CommonJS and ES6 code (for example, any node module does
it).  But, though it is internally the same structure, I agree it is
not conceptually the same as exporting explicitly the same thing under
two different names in ES6.

wincent:
> There are some libraries that export the same thing both as module.exports.default and module.exports to ease interoperability between CommonJS and ES6 code (for example, any node module does it). But, though it is internally the same structure, I agree it is not conceptually the same as exporting explicitly the same thing under two different names in ES6.

And I would hope that if they do this, it's a build thing and they're not actually slavishly writing it out by hand.

slaves

izaera:
> And I would hope that if they do this, it's a build thing and they're not
> actually slavishly writing it out by hand.
> 
I have no idea. It is trivial to write a Babel plugin that does it, but who
knows...

izaera:
I have removed the double exports

izaera:
@wincent I have removed the double exports

izaera:
@wincent I have removed the duplicated exports

izaera:
@wincent I have exported things as default where it applies

izaera:
etc. :smiling_imp:

PR #576, opened 2020-05-18
feat: #522 implement Angular adaptation

izaera:
This PR implements Angular adaptation as in bundler 2 and then refactors both the React and Angular adaptations to generalize them, given that all adaptations should basically share the same code, as the operations are similar in spirit, though they may vary a bit in their implementation details.

I have tested both adaptations locally with the portlets in the QA dir and they work for the very basic setup. We will need more QA testing before release, but I think it is enough for merging.


wincent:
I think this is targeting the wrong branch. It shows 162 commits and a bunch of merge conflicts.

Screenshot 2020-05-18 -180530-oCTsNPeY@2x

I'm going to see if it lets me toggle it to target `3.x-WIP`.

packages/liferay-npm-bundler/src/steps/adapt/angular-cli/index.ts
```diff
@@ -0,0 +1,70 @@
...
+	copyStaticAssets,
+	processAdapterModules,
+	processCssFiles,
+	processPackageJson,
+	processWebpackBundles,
+} from '../index';
+import adaptStaticURLsAtRuntime from '../transform/js/operation/adaptStaticURLsAtRuntime';
+import tweakAttachmentToDOM from './tweakAttachmentToDOM';
+
+const {pkgJson} = project;
```
wincent, line 22: So we do quite a bit of work while traversing the (internal) module graph of this package, which effectively creates hidden global state which might limit our flexibility in the future. For example, `project` here comes into being at module level here as a result of the `new Project('.')` that happens in `globals.ts`, and it happens during module initialization. That means that we can never adapt more than one project at a time, and it might make testing harder (although Jest it will blow away module state between tests, it's still awfully magical).

I get that this all works right now, so I'm not saying you should change it, but I just wanted to point out the perils of implicit global state (contrasted with the explicit global state, which you could get via an explicit, lazy call — eg. a `getGlobals()` or something).

packages/liferay-npm-bundler/src/steps/adapt/angular-cli/tweakAttachmentToDOM.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param domId DOM node id (f.e.: 'app-root')
+ */
+export default function tweakAttachmentToDOM(domId: string): JsSourceTransform {
+	return (async (source) =>
+		await replaceJsSource(source, {
+			enter(node) {
+				if (
+					node.type !== 'Literal' ||
+					typeof node.value !== 'string' ||
```
wincent, line 24: Do you need this runtime check? `domId` is statically known to be a `string`, so the `node.value !== domId` would fail anyway if `typeof node.value` were not `'string'`.

packages/liferay-npm-bundler/src/steps/adapt/index.ts
```diff
@@ -7,38 +7,84 @@ import fs from 'fs-extra';
...
 import * as log from '../../log';
-import {findFiles} from '../../util/files';
+import {copyFiles, findFiles} from '../../util/files';
 import Renderer from '../../util/renderer';
 import exportModuleAsFunction from './transform/js/operation/exportModuleAsFunction';
 import namespaceWepbackJsonp from './transform/js/operation/namespaceWepbackJsonp';
+import replace from './transform/text/operation/replace';
+
+/**
+ * Descripition of framework's webpack build output so that adapted modules can
```
wincent, line 28: Descripition → Description

packages/liferay-npm-bundler/src/steps/adapt/index.ts
```diff
@@ -7,38 +7,84 @@ import fs from 'fs-extra';
...
 /**
  * Generate adapter modules based on templates.
  *
  * @param data extra values to pass to render engine in addition to `project`
  */
-export async function processAdapterModules(data: object): Promise<void> {
-	const renderer = new Renderer(
+export async function processAdapterModules(
+	webpackBundles: WebpackBundles,
+	data: object = {}
```
wincent, line 69: `object` is a pretty weak type; can we say anything more specific here?

packages/liferay-npm-bundler/src/steps/adapt/index.ts
```diff
@@ -100,6 +147,63 @@ async function processAdapterModule(
...
+				sourceAsset.asPosix,
+				`o${project.jar.webContextPath}/${sourceAsset.asPosix}`
+			);
+
+			return map;
+		},
+		new Map<string, string>()
+	);
+
+	await Promise.all(
```
wincent, line 185: Obligatory comment about unbounded/unbatched use of `Promise.all` (yeah, yeah, I know).

packages/liferay-npm-bundler/src/steps/adapt/index.ts
```diff
@@ -100,6 +147,63 @@ async function processAdapterModule(
...
+		cssFiles.map(async (file) => {
+			await transformTextFile(
+				adaptBuildDir.join(file),
+				buildBundlerDir.join(file),
+				replace(assetURLsMap)
+			);
+		})
+	);
+
+	log.debug(`Processed ${cssFiles.length} CSS files`);
```
wincent, line 195: Pro-Tip™: you can avoid the eye-sore of printing "1 CSS files" and still not have to implement single/plural handling by rewording this:

```js
log.debug(`Processed CSS files: ${cssFiles.length}`);
```

packages/liferay-npm-bundler/src/steps/adapt/index.ts
```diff
@@ -100,6 +147,63 @@ async function processAdapterModule(
...
+}
+
+/**
+ *
+ * @param cssPortletHeader
+ * Path to CSS file to use as header portlet CSS file.
+ *
+ * Note that it may contain `[filename]` when the CSS is stored in a file with a
+ * webpack hash in its name. In that case, the `[filename]` will be replaced by
+ * `filename.hash`.
```
wincent, line 205: Are these comments about hashes still accurate? (I know you removed some fancy hash-matching code in a previous PR.)

packages/liferay-npm-bundler/src/steps/adapt/index.ts
```diff
@@ -111,7 +215,81 @@ export async function processPackageJson(
...
+	} else if (candidateFiles.length > 1) {
+		throw new Error(
+			`Framework's build produced more than one ${file} artifact: ` +
+				`cannot determine which one to use`
+		);
+	}
+
+	const filename = candidateFiles[0].basename().toString();
+
+	const extRegexp = new RegExp(`${escapeStringRegexp(path.extname(file))}$`);
```
wincent, line 278: Nit: because `RegExp`, this should be `extRegExp`, not `extRegexp` (and we could say the same about `escapeStringRegexp`).

packages/liferay-npm-bundler/src/steps/adapt/angular-cli/index.ts
izaera: I know, I know. This is like that for "historical reasons" (I love that). In the beginning, there was not `Project` class at all and everything was spread here and there. When it became unmanageable, I decided to create the `Project` class to at least centralize the behavior in one place. However, because the code was already there and couldn't be changed entirely by that time, I had to create it as a global variable.

Then I added the `dir` properties to `Project` to make any reference to the project directory relative to them and thus be able to, for example, work with a project in a directory far from the  `CWD`. That also made explicit where things where supposed to be. And it also let me do tests by using `new Project` in those tests where I could pass the `Project` object to the SUT. In tests where the `Project` was used as a global I had to resort to the `loadFrom` method (not very clean, but worked; though imposing the need to serialize tests, not allowing to run them in parallel).

This time, I have removed the `project` global from `js-toolkit-core` so as to avoid it being part of the API (at least) and have hidden the global inside the bundler (which is not public API). With the idea of evolving this more into a solution where several `Project` objects can be used at the same time, as you propose.

However, don't feel very happy, because every other auxiliary tool we use to make the builds (mainly webpack and its plugins) assume that everything happens relative to CWD (remember our initial implementation I talked about? :sweat_smile: ) so even if we allow our code to deal with more than one Project we will still have the limitation of the tools we use :-(.

Also, even though I agree it is better to make the global state explicit, I don't like to have a `getGlobals()` call everywhere because it makes the code more complex. We had some other projects where that pattern was used and it was nearly impossible to know where things were and where they came from; even though the pattern is supposed to make it more explicit. This is because people sometimes passed the globals as parameters in functions, in other places they called the `getGlobals` equivalent on top of the module, other times they called it inside the functions, etc... If you are curious about what project, it was in the Themes Toolkit :scream: .

Having a single import at the top seems to help a bit and if some day we need to have more than one Project instance we can make use of a bit of magic and inject the necessary instance by using Zones and a Proxy for example (or maybe a getter in `module.exports`, etc...). 

I'm not saying that we will ever do it (because of the CWD limitation), but given that `project` is the one and only implicit global in any bundler process, I prefer the simplicity that a global brings to all the code even if I have to use some little magic to inject it in the code.

packages/liferay-npm-bundler/src/steps/adapt/index.ts
izaera: No, we can't because the values there are ultimately passed to a `.ejs` template, so the structure of the object depends on the template :shrug: .

It's a bad thing and it's a place for future failures, for sure, but I can't see a better way to do it  without making it more complex. 

A solution could be to have one function per template which typed the object we are passing. Do you think it would be worth to make the code more complex for the sake of typing the object we pass to the template? :thinking: I like the idea, but I don't know if it will be much better...

packages/liferay-npm-bundler/src/steps/adapt/index.ts
izaera: :sweat_smile: . I'll add an issue to review all these parallel actions in windows at the end so that we don't forget about it :+1:

izaera: done -> https://github.com/liferay/liferay-js-toolkit/issues/577

packages/liferay-npm-bundler/src/steps/adapt/index.ts
izaera: OK. Bundler 2 removed the hashes from filenames because of its limitations when defining rules to select files inside the project build. We used static regexps that couldn't account for unknown hashes so we needed to remove them before applying the processes.

Now we don't have that limitation, so I modified that code in a previous PR to avoid dealing with hash removals. 

So now the situation is that depending on the framework you are adapting the files will or won't have hashes. For example: React uses hashes, Angular doesn't. 

We have to take care of hashes but we don't need to remove them any more. This makes development easier for us (though it seems counter-intuitive) and for our users because the files we see in our filesystems are the files that end up in the JAR and the files we get over HTTP, so we don't need to mentally convert them depending on the context.

packages/liferay-npm-bundler/src/steps/adapt/index.ts
izaera: :scream: :+1:

packages/liferay-npm-bundler/src/steps/adapt/angular-cli/tweakAttachmentToDOM.ts
izaera: I guess it can be removed, yes.

packages/liferay-npm-bundler/src/steps/adapt/angular-cli/index.ts
izaera: One thing to note here is that by doing `const {pkgJson} = project;` I have made what I said was not a good idea :sweat_smile: because that way you don't know where things come from.

wincent: > Also, even though I agree it is better to make the global state explicit, I don't like to have a `getGlobals()` call everywhere because it makes the code more complex. We had some other projects where that pattern was used and it was nearly impossible to know where things were and where they came from; even though the pattern is supposed to make it more explicit. This is because people sometimes passed the globals as parameters in functions, in other places they called the `getGlobals` equivalent on top of the module, other times they called it inside the functions, etc... If you are curious about what project, it was in the Themes Toolkit 😱 .

I don't think this is an effective argument against explicit global state. This is just pointing at an example of poorly done global state and claiming that that proves that global state is better off being implicit.

But I'm not trying to talk you into making any changes — I'm just stating an opinion.

packages/liferay-npm-bundler/src/steps/adapt/index.ts
wincent: Seems like it is probably not worth it. In principle, we can always tighten things up later on.

I've seen other projects where people say "we can always tighten things up later on" and it actually ends up being suuuuper difficult because some weak assumption has ended up spreaded over the entire program, and you make a change to a type in one place (ie. not a code/functionality change, you just try to get rid of an `any`) and it ends up creating an avalanche of cascading type errors that are very hard to fix.

I'm not to worried about it in this case though because the boundary between the template and the rest of the code seems to be pretty limited.

izaera: Yes, I think that's the case. The templates are something that usually are run locally so I don't expect their contexts to be passed across several places of the code.

PR #582, opened 2020-05-26
feat: portal build integration

izaera:
This PR adds miscellaneous features and fixes needed to integrate liferay-npm-bundler into the Portal's build.

It has been tested locally and we need to release a new alpha version once it is merged.

Please refer to each commit for a description of the modifications.


CONTRIBUTING.md
```diff
@@ -103,15 +103,53 @@ Both must be deployed to see them in action.
...
+# If any checks fail, fix them, submit a PR, and when it is merged,
+# start again. Otherwise...
 
-The script performs the necessary checks before release and then publishes a new version. It doesn't leave any trace in git because the version number contains the commit hash.
+# Change to the directory of the package you wish to publish:
+cd packages/liferay-npm-bundler
 
-Even though the script checks for a clean working copy, sometimes it may be necessary to tweak the `package.json` of the released package to point it to another dependency snapshot. In such cases you just need to modify the `package.json` file locally and the script won't complain about that situation.
+# Update the changelog:
+npx liferay-changelog-generator --version=liferay-npm-bundler/v3.0.0
```
wincent, line 123: Pro-Tip™: as of https://github.com/liferay/liferay-npm-tools/pull/430 you don't need to spell out `liferay-npm-bundler`, *provided* that you have a package-local `.yarnrc` file with a `version-tag-prefix` in it.

Of course, you don't in this project but maybe one day you will...

CONTRIBUTING.md
```diff
@@ -103,15 +103,53 @@ Both must be deployed to see them in action.
...
+## Releasing local-only versions
+
+If you need to test local versions of the packages, you can install [Verdaccio](https://verdaccio.org) (a local NPM repository). Verdaccio is usually located at `http://localhost:4873` and you can use these commands to work with it:
+
+1. To publish a local-only version set the desired `package.json` to the new version number and run `npm publish --registry http://localhost:4873`.
+2. To use the local repository from `npm` run `npm set registry http://localhost:4873`.
+3. To use the local repository from `yarn` run `yarn config set registry http://localhost:4873`.
+4. To stop using the local repository edit your `~/.npmrc` and `~/.yarnrc` files and remove the local repo.
+
+Publishing to the local Verdaccio repository won't update [https://npmjs.com](https://npmjs.com) so you can publish as much local versions as you want without worrying about polluting the public npm repository. Then, when you are finished testing, just remove the local versions from you local Verdaccio, point `npm` and `yarn` to the public npm repo, and publish the ultimate valid release.
```
wincent, line 155: "as much" → "as many"

packages/liferay-js-toolkit-core/src/project/Project.ts
```diff
@@ -157,6 +158,20 @@ export default class Project {
 		return this._sources;
 	}
 
+	get srcDir(): FilePath {
+		if (this._srcDir === undefined) {
+			let dir = prop.get(this._configuration, 'source', '.');
+
+			if (dir !== '.' && !dir.startsWith('./')) {
```
wincent, line 165: What if it starts with `../`? eg. given "../foo", we'll wind up with `dir = './../foo'`, which is admittedly harmless, but the fact that the `if` special-cases exactly one kind of relative path is distracting.

packages/liferay-npm-bundler/src/steps/bundle/adapt/write-export-modules.ts
```diff
@@ -175,8 +175,14 @@ async function writeLocalExportModule(
 		.replace('./', '')
 		.replace(/\.js$/, '')}`;
 
-	const bundlesLocation = moduleFile.dirname().relative(buildBundlerDir)
+	let bundlesLocation = moduleFile.dirname().relative(buildBundlerDir)
 		.asPosix;
 
+	if (bundlesLocation === '') {
+		bundlesLocation = '.';
+	} else if (!bundlesLocation.startsWith('./')) {
```
wincent, line 183: Same thing here; not handling `../`-type relative paths is a bit distracting. Would be nice to have an `isRelative()` function to abstract this away.

packages/liferay-npm-bundler/src/steps/bundle/ExplainedError.ts
```diff
@@ -0,0 +1,93 @@
...
+		filePath = '<unknown file>';
+	}
+
+	const contextLines =
+		sourceCode && line ? getContextLines(sourceCode, line) : '';
+
+	return `${error.toString()}${contextLines}\n...at ${filePath}`;
+}
+
+function pad(minChars: number, value: number): string {
```
wincent, line 85: What version of Node do we have to support here? [`String.prototype.padStart`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) exists in Node v8 and up.

packages/liferay-npm-bundler/src/steps/bundle/plugin/imports-loader.ts
```diff
@@ -7,16 +7,18 @@ import {parse} from 'acorn';
...
 import estree from 'estree';
-import {FilePath, Project, addNamespace} from 'liferay-js-toolkit-core';
+import {Project, addNamespace} from 'liferay-js-toolkit-core';
 
 import {project} from '../../../globals';
 import report from '../../../report';
 import ReportLogger from '../../../report/logger';
 
 export default function (content: string): string {
+	const prjDirRelResourceFile = project.dir.relative(this.resourcePath);
```
wincent, line 17: We [should probably avoid abbreviations](https://github.com/liferay/liferay-frontend-guidelines/issues/159).

packages/liferay-npm-bundler/src/steps/bundle/plugin/imports-loader.ts
```diff
@@ -44,19 +46,33 @@ export default function (content: string): string {
...
-	log.info(`File contained 'imports' which were diverted to runtime loader`);
+		return generate(ast, {});
+	} catch (err) {
+		// If we've been unable to parse the file it's because it is not valid
+		// JavaScript so there shouldn't be any import in there.
+		//
+		// Also, it is almost certain that webpack will fail after us when
+		// parsing the same content, so we leave the error handling to webpack
+		// better to avoid user confusion.
+		log.error(`File could not be parsed`, err);
```
wincent, line 72: This might be the thing that is missing from the commit message, which does a good job of explaining *what* change but doesn't really say *why*:

```
Move ExplainedError to its own file and stop using it in imports-loader.
Instead, log parse errors in imports-loader to the report file and don't fail
the build.
```

Is this comment the "why"?

packages/liferay-js-toolkit-core/src/project/Project.ts
izaera: Mmm, source dir should never start with '..'  as that would point outside the project folder :scream: . I'd rather check and throw an error in that case.

packages/liferay-npm-bundler/src/steps/bundle/ExplainedError.ts
izaera: Didn't know that method existed. We will support node LTS or higher, so it is safe to use it.

packages/liferay-npm-bundler/src/steps/bundle/plugin/imports-loader.ts
izaera: I began to use those abbreviations to know what each path variable contains. And decided to use small strings because the nomenclature can explode easily. For example you may end up having variables like `projectDirRelativeImageFilePosixPath` :scream: .

Said that, it may be well possible that we rename/refactor all these variables in the future making use of `FilePath` so that it is clear what they are. For example, if we add the `isRelative()` method you propose above and provide access to relative and absolute paths to the most relevant project folders as I propose in a TODO in this PR, it is possible that we don't need this kind of variables any more.

packages/liferay-npm-bundler/src/steps/bundle/plugin/imports-loader.ts
izaera: Yes. I will add that to the commit instead. I made two edits and now the commentary doesn't make sense because we already call `log.error` (in the first edit there was the comment without the log).

packages/liferay-npm-bundler/src/steps/bundle/plugin/imports-loader.ts
wincent: Well, in this very PR, you add `buildGeneratedDirRelativeModuleFile`, so it seems like you're not *always* worried about long names.

FWIW, I like simple rules that easily explained and applied consistently — eg. I prefer a rule like "don't abbreviate" to a rule like "try not to use abbreviations, except in cases where the variable can get too long".

And as much as I hate exceptions, I think its OK to use abbreviations when there is some "standard" abbreviation in usage by strong convention (eg. `i` for loop variables) or due to a technical constaint (eg. to avoid a reserved word, like `pkg` instead of reserved `package`) or for consistency with an abbreviation coming from "outside" (eg. using a Node API like `fs.readdir`, it's natural to keep using `dir` instead of spelling out `directory`; or calling something `someRegExp` because the built-in class name is `RegExp` and not `RegularExpression`). But I normally try to leave out discussion of these because normally the simple explanation is good enough.

PR #592, opened 2020-06-03
fix: #588 enhance error detection on dynamic requires

### LINKED ISSUE #588: Bundler failing with `undefined` dependency when a dynamic reference exists in a package
Trying to use react-live module but with no success, I noticed that the bundler was creating an undefined package in one of the dependencies which was causing the react widget to fail. The widget was created using `yo liferay-bundle`.

```
Liferay.Loader.define('my-project$regexpu-core@4.7.0/rewrite-pattern', ['module', 'exports', 'require', 'my-project$regjsgen', 'my-project$regjsparser', 'my-project$regenerate', 'my-project$unicode-match-property-ecmascript', 'my-project$unicode-match-property-value-ecmascript', './data/iu-mappings', './data/character-class-escape-sets', 'my-project$undefined'], function (module, exports, require) {
...
```
Notice `my-project$undefined` at the end.

Browser log error:
```
Some information about the require() call follows:
   · Require call id: 1 
   · Required modules: ["my-project@1.0.0"] 
   · Missing modules: (n/a) 
   · Stack trace of the require() call: 
        Error
            at e.value (http://localhost:8080/o/frontend-js-web/loader/loader.js:1:5898)
 Error: The following problems where detected while resolving modules:
    · Missing dependency ':ERROR:Missing version constraints for my-project$undefined in package.json of my-project$regexpu-core@4.7.0' of 'my-project$regexpu-core@4.7.0/rewrite-pattern'
```

And this is from liferay-npm-bundler-report:
```
Failed to resolve dependency my-project$undefined with error: Error: Cannot find module 'undefined/package.json' from './node_modules/regexpu-core'
```

`.npmbuildrc`:
```
{
	"translatorTextKey": "",
	"supportedLocales": [],
	"liferayDir": "/Users/julienmourad/projects/bundles/liferay-dxp-7.2.10-dxp-4",
	"webpack": {
		"rules": [
			{
				"test": "src\\/.*\\.js$",
				"use": "babel-loader"
			}
		]
	}
}
```

`.npmbundlerrc`:
```
{
	"create-jar": {
		"output-dir": "dist",
		"features": {
			"js-extender": true,
			"web-context": "/my-project",
			"localization": "features/localization/Language",
			"configuration": "features/configuration.json"
		}
	},
	"dump-report": true
}
```

`package.json`:
```
"devDependencies": {
	"liferay-npm-bundler": "^2.8.0",
	"liferay-npm-build-support": "^2.8.0",
	"copy-webpack-plugin": "4.6.0",
	"webpack": "4.29.6",
	"webpack-cli": "3.3.0",
	"webpack-dev-server": "3.2.1",
	"babel-cli": "6.26.0",
	"babel-preset-env": "1.7.0",
	"babel-preset-react": "6.24.1",
	"babel-loader": "7.1.5"
},
"dependencies": {
	"react": "16.8.6",
	"react-dom": "16.8.6",
	"react-live": "^2.2.2"
},
"scripts": {
	"build": "babel --source-maps -d build src && npm run copy-assets && liferay-npm-bundler",
	"copy-assets": "lnbs-copy-assets",
	"translate": "lnbs-translate",
	"deploy": "npm run build && lnbs-deploy",
	"start": "lnbs-start"
}
```

izaera:
NOTE: THIS DOES NOT FIX #588 ENTIRELY, JUST MAKES IT A BIT LESS PROBLEMATIC

I have added code to ignore dynamic requires during build time. Now we don't
generate the undefined dependency, just skip it and let it fail during execution
of the module (note that it may even not fail if the code flow doesn't make the
dynamic require to be executed).

Also, an error is written to the report file.

This should alleviate the problem because:

1. It may make some setups work (as long as they don't execute the dynamic
   require)
2. If they execute the require, a better error will be shown in the console
3. The problem is correctly explained in the report file

I have tested the PR manually with the package involved in issue #588

### DIFF
--- a/packages/babel-plugin-wrap-modules-amd/src/index.js
+++ b/packages/babel-plugin-wrap-modules-amd/src/index.js
@@ -53,11 +53,19 @@ export default function({types: t}) {
 				) {
 					const argument0 = parent.arguments[0];
 
-					if (t.isLiteral(argument0)) {
-						const moduleName = argument0.value;
+					if (argument0.type !== 'StringLiteral') {
+						log.error(
+							'wrap-modules-amd',
+							'Module has a non static require, which is not ' +
+								'supported: module may fail when executed'
+						).linkToIssue(588);
 
-						dependencies[moduleName] = moduleName;
+						return;
 					}
+
+					const moduleName = argument0.value;
+
+					dependencies[moduleName] = moduleName;
 				}
 			}
 		},
--- a/packages/babel-plugin-wrap-modules-amd/src/index.js
+++ b/packages/babel-plugin-wrap-modules-amd/src/index.js
@@ -47,26 +47,39 @@ export default function({types: t}) {
 				}
 
 				if (
-					t.isCallExpression(parent) &&
-					parent.callee === node &&
-					parent.arguments.length == 1
+					!t.isCallExpression(parent) ||
+					parent.callee !== node ||
+					parent.arguments.length != 1
 				) {
-					const argument0 = parent.arguments[0];
+					return;
+				}
 
-					if (argument0.type !== 'StringLiteral') {
-						log.error(
-							'wrap-modules-amd',
-							'Module has a non static require, which is not ' +
-								'supported: module may fail when executed'
-						).linkToIssue(588);
+				const argument0 = parent.arguments[0];
 
-						return;
-					}
-
+				if (argument0.type === 'StringLiteral') {
 					const moduleName = argument0.value;
 
 					dependencies[moduleName] = moduleName;
+
+					return;
 				}
+
+				if (
+					argument0.type === 'TemplateLiteral' &&
+					argument0.quasis.length === 1
+				) {
+					const moduleName = argument0.quasis[0].value.raw;
+
+					dependencies[moduleName] = moduleName;
+
+					return;
+				}
+
+				log.error(
+					'wrap-modules-amd',
+					'Module has a non static require, which is not ' +
+						'supported: module may fail when executed'
+				).linkToIssue(588);
 			}
 		},
 	};


packages/babel-plugin-wrap-modules-amd/src/index.js
```diff
@@ -53,11 +53,19 @@ export default function({types: t}) {
 				) {
 					const argument0 = parent.arguments[0];
 
-					if (t.isLiteral(argument0)) {
-						const moduleName = argument0.value;
+					if (argument0.type !== 'StringLiteral') {
```
wincent, line 56: It could also be a template literal, which would only be an error if it contained interpolations.

izaera: :thinking: Mmm, that's true, although we don't support template literals because our input must be CommonJS/ES5.

In the offending module (`regexpu-core/rewrite-pattern`) everything is transpiled to CommonJS/ES5 except for the dynamic require that contains a template literal, probably to make use of the [webpack dynamic require interpolation feature](https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import) because it says:

```javascript
require(`regenerate-unicode-properties/${ path }.js`);
```
I agree it wouldn't harm to support template literals even if we require CommonJS/ES5 as input, but it is more complex than it seems at first glance and we have never encountered any static require with template literals in the wild. We know this since we didn't support it before this PR and nobody has ever reported a bug like that. 

So, given that this is a fix to just show better errors and that it doesn't even fix the root cause of the problem, I think it is not worth spending more time on it to support a condition that we don't even know if will happen...

wincent: > but it is more complex than it seems at first glance

Perhaps. I think for anything but pathological cases it is trivial, at least it seemed that way when I've done it in the past ([example](https://github.com/liferay/eslint-config-liferay/blob/1720ded29a837eeef2aa1ac57b2236fcc6fb7d1a/plugins/eslint-plugin-liferay/lib/rules/import-extensions.js#L37-L43)).

Issue #596, opened 2020-06-10
Support 'global' as in Node.js


izaera:
Some polyfilll modules inspect the `global` variable to look for definitions. Those module would fail in an AMD environment, but if we point `global` to `window` (similar to what webpack does) some of them may work seamlessly.

Appeared with one dependency of package `string.prototype.trimright` while testing the fix for #589.

Issue #589, opened 2020-06-01
Babel plugin babel-plugin-namespace-amd-define fails when wrapping modules that contain a `define` variable


aggie2016:
When building a project that contains polyfills (or any npm_module that declares a variable called "define", the `babel-plugin-namespace-amd-define` doesn't properly prepend `Liferay.Loader`.

See below original source code:
```
'use strict';

var callBind = require('es-abstract/helpers/callBind');
var define = require('define-properties');

var implementation = require('./implementation');
var getPolyfill = require('./polyfill');
var shim = require('./shim');

var bound = callBind(getPolyfill());

define(bound, {
	getPolyfill: getPolyfill,
	implementation: implementation,
	shim: shim
});

module.exports = bound;

```

Source code after `babel-plugin-namespace-amd-define` finishes running:
The final packaged source contains a redefinition of `define`.
```
'use strict';

define('digital-field-guide-app$string.prototype.trimright@2.1.2/index', ['module', 'exports', 'require', 'digital-field-guide-app$es-abstract/helpers/callBind', 'digital-field-guide-app$define-properties', './implementation', './polyfill', './shim'], function (module, exports, require) {
	var define = undefined;
	var callBind = require('digital-field-guide-app$es-abstract/helpers/callBind');
	var define = require('digital-field-guide-app$define-properties');

	var implementation = require('./implementation');
	var getPolyfill = require('./polyfill');
	var shim = require('./shim');

	var bound = callBind(getPolyfill());

	define(bound, {
		getPolyfill: getPolyfill,
		implementation: implementation,
		shim: shim
	});

	module.exports = bound;
});
//# sourceMappingURL=index.js.map
```

PR #598, opened 2020-06-10
feat: #596 support 'global' object

izaera:
Simply define a `global` object pointing to `window` like webpack does so that
some modules work without problems.

Tested with package `string.prototype.trimright` of #589.

### DIFF
--- a/packages/babel-plugin-wrap-modules-amd/src/__tests__/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin-wrap-modules-amd/src/__tests__/__snapshots__/index.test.js.snap
@@ -3,6 +3,7 @@
 exports[`correctly wraps modules 1`] = `
 "define(['module', 'exports', 'require'], function (module, exports, require) {
 	var define = undefined;
+	var global = window;
 
 	console.log('Say something');
 	if (1 == 0) {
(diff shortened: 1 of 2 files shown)


izaera:
@wincent The truth is I don't have any idea of where it's done. I just built a webpack bundle for a dependency using `global` and saw the resulting code which passes `global` as a parameter pointing to the IIFE of the webpack bundle (not `window`). 

However, because we don't have any IIFE, the thing that most resembles it is, probably, `window`.

Let me check if I can find any link that shows evidence on that...

izaera:
I couldn't find where it does the thing, but here's the code it generates :point_down: 

I have an `index.js` file like this:

```javascript
console.log(global);
```

which webpacks transforms to:

```javascript

...

/***/ "./index.js":
/*!******************!*\
  !*** ./index.js ***!
  \******************/
/*! no static exports found */
/***/ (function(module, exports, __webpack_require__) {

/* WEBPACK VAR INJECTION */(function(global) {console.log(global);

/* WEBPACK VAR INJECTION */}.call(this, __webpack_require__(/*! ./node_modules/webpack/buildin/global.js */ "./node_modules/webpack/buildin/global.js")))

/***/ }),

/***/ "./node_modules/webpack/buildin/global.js":
/*!***********************************!*\
  !*** (webpack)/buildin/global.js ***!
  \***********************************/
/*! no static exports found */
/***/ (function(module, exports) {

var g;

// This works in non-strict mode
g = (function() {
        return this;
})();

try {
        // This works if eval is allowed (see CSP)
        g = g || new Function("return this")();
} catch (e) {
        // This works if the window reference is available
        if (typeof window === "object") g = window;
}

// g can still be undefined, but nothing to do about it...
// We return undefined, instead of nothing here, so it's
// easier to handle this case. if(!global) { ...}

module.exports = g;


/***/ })

...

```

In fact, I lied because it does use the `window` object as `global` (like us).

wincent:
Thanks. It's useful to have that documented somewhere.

Issue #600, opened 2020-06-15
Warning when deploying jar due to possible duplicate of Provide-Capability in Manifest


ricardocousolr:
When reading the MANIFEST.MF from the jar file obtained by executing `npm run build:liferay` (or similar), the `java.util.jar.Attributes` class can issue the following warning:
```
java.util.jar.Attributes.read Duplicate name in Manifest: Provide-Capability.
Ensure that the manifest does not have duplicate entries, and
that blank lines separate individual sections in both your
manifest and in the META-INF/MANIFEST.MF entry in the jar file.
```

This happens becase the MANIFEST.MF includes the repeated key `Provided-Capability`:
```
Manifest-Version: 1.0
Bundle-ManifestVersion: 2
Tool: liferay-npm-bundler-2.18.5
Bundle-SymbolicName: barebones-app
Bundle-Version: 0.1.2
Web-ContextPath: /barebones-app-0.1.2
Provide-Capability: osgi.webresource;osgi.webresource=barebones-app;version:Version="0.1.2"
Provide-Capability: liferay.resource.bundle;resource.bundle.base.name="content.Language"
Require-Capability: osgi.extender;filter:="(&(osgi.extender=liferay.frontend.js.portlet)(version>=1.0.0))"
```
When reading the manifest, the only the latest value will be considered, so the warning is issue to inform the user. 

The relevant code is modify would be https://github.com/liferay/liferay-js-toolkit/blob/master/packages/liferay-npm-bundler/src/jar/index.js#L120-L132

(This issue can affect customers; e.g. [LPP-37452](https://issues.liferay.com/browse/LPP-37452))

PR #601, opened 2020-06-16
fix: #600 avoid duplicate entries in JAR manifest

izaera:
Tested manually (inspecting the JAR file and asserting the portlet localization works when deployed to Liferay).

### DIFF
--- a/packages/liferay-npm-bundler/src/jar/manifest.ts
+++ b/packages/liferay-npm-bundler/src/jar/manifest.ts
@@ -5,6 +5,18 @@
 
 import project from 'liferay-npm-build-tools-common/lib/project';
 
+const FORBIDDEN_CUSTOM_HEADERS = new Set([
+	'Bundle-ManifestVersion',
+	'Bundle-Name',
+	'Bundle-SymbolicName',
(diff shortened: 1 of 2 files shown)


packages/liferay-npm-bundler/src/jar/manifest.ts
```diff
@@ -0,0 +1,115 @@
...
+		'Bundle-ManifestVersion',
+		'Bundle-Name',
+		'Bundle-SymbolicName',
+		'Bundle-Version',
+		'Manifest-Version',
+		'Provide-Capability',
+		'Require-Capability',
+		'Tool',
+		'Web-ContextPath',
+	];
```
wincent, line 93: What is the point of this being a static member? This whole file looks more like Java than JS. 

I'd just make this a `const FORBIDDEN_CUSTOM_HEADERS = new Set(['blah', 'etc'])` and hide it in the module closure. (Using `Set` to avoid `O(n)` lookup when you do your `includes()` call.)

izaera: :+1:

Issue #604, opened 2020-06-18
babel-polyfill fails with TypeError: "define is not a function"


izaera:
The module `babel-polyfill@6.26.0/lib/index.js` has the following code:

```javascript
  function define(O, key, value) {
    O[key] || Object[DEFINE_PROPERTY](O, key, {
      writable: true,
      configurable: true,
      value: value
    });
  }

  define(String.prototype, "padLeft", "".padStart);
  define(String.prototype, "padRight", "".padEnd);
```

but because, when wrapping the modules for AMD loader, we add this on top of the module:

```javascript
  var define = undefined;
```

the JS engine cannot find `define`.

Note that this is a very exotic case because `var define = undefined;` covers all cases except for this one. 

In effect, if `define` is already present at the global scope, `var define = undefined;` masks it. 

And in the case of modules that define a `define` variable, everything is OK because if you have this code: 

```javascript
  var define = undefined;
  console.log(define);
  var define = 'perico';
  console.log(define);
```

you get:

```
undefined
perico
```

as expected, and the JS engine doesn't complain about duplicate definitions.

However, it seems that in the case of variable vs function, the variable wins and thus `babel-polyfill` fails.

izaera:
It seems that it's not that in variable vs function variable wins, but in this case, because of hoisting one is overwriting the other.

See https://stackoverflow.com/questions/336859/var-functionname-function-vs-function-functionname#comment16104094_336859

izaera:
Given that it is a hoisting problem, doing this (putting the module code in a block after the `var define = undefined;`) solves the issue:

```javascript
  var define = undefined;
  
  {
    function define(O, key, value) {
      O[key] || Object[DEFINE_PROPERTY](O, key, {
        writable: true,
        configurable: true,
        value: value
      });
    }

    define(String.prototype, "padLeft", "".padStart);
    define(String.prototype, "padRight", "".padEnd);
  }
```

I can't think of any pitfall in this construct, but I say that same sentence every time I touch this... :shrug:

wincent:
> I can't think of any pitfall in this construct

That makes two of us. About the only thing I can imagine is some overzealous minifier deciding that you didn't need the block and stripping it out.

izaera:
:thinking: That shouldn't happen during the build because the bundler is usually the last step to run. Though you could theoretically configure such a setup.

However, we have a minifier in the portal that could do it.

However, if it did we would have seen it already, as the previous code was subject to that error too.

[...]

PR #605, opened 2020-06-18
fix: #604 put module definition in a block scope

izaera:
This is to avoid the variable declarations on top of the module definition be
mixed with module's ones because of hoisting.

See the #604 issue for a more detailed explanation of this commit.


izaera:
Going to merge this as it has already been discussed in the associated issue.

Issue #550, opened 2020-03-27
Decide final configuration API for exports


izaera:
Currently we have `imports` and `exports`. 

But webpack has `entry`, very similar to  `exports`.

What shall we use?

izaera:
We still use `imports` because it was called like that in Bundler 2, but as we are developing a new version, we can change it.

However, we should not deviate from Bundler 2 more than necessary because it will make migration more difficult. 

Though in the case of `imports` you need to migrate the config anyway by moving it to a different section, so the developer could as well rename it in the same step.

jbalsas:
So, in general terms, I'd like to reuse existing, well-defined concepts when possible instead of having to cook our own.

### Exports

This seems analog to [webpack Entry Points](https://webpack.js.org/concepts/entry-points). It supports:
- [Single Entry (Shorthand) Syntax](https://webpack.js.org/concepts/entry-points/#single-entry-shorthand-syntax): `entry: string | [string]`
- [Object Syntax](https://webpack.js.org/concepts/entry-points/#object-syntax): `entry: { <entryChunkName> string | [string] }`

### Imports

This might be similar to [webpack Module Resolution](https://webpack.js.org/concepts/module-resolution/) which uses the [Resolve API](https://webpack.js.org/configuration/resolve/#resolve). This is also somewhat connected with [Yarn Resolutions](https://classic.yarnpkg.com/en/docs/package-json#toc-resolutions) in the sense that regardless what a `package.json` says, you're giving an app a different dependency. Bits of this API that would be helpful:

#### [`resolve.alias`](https://webpack.js.org/configuration/resolve/#resolvealias)

```javascript
  resolve: {
    alias: {
      Utilities: path.resolve(__dirname, 'src/utilities/'),
      Templates: path.resolve(__dirname, 'src/templates/')
    }
  }
```

#### [`resolve.mainFields](https://webpack.js.org/configuration/resolve/#resolvemainfields)

```javascript
  resolve: {
    mainFields: ['browser', 'module', 'main']
  }
```


#### [`resolve.mainFiles](https://webpack.js.org/configuration/resolve/#resolvemainfiles)

```javascript
  resolve: {
    mainFiles: ['index']
  }
```

We shouldn't need to implement the whole `resolve` API... and I am afraid of running straight into the [Uncanney Valley](https://en.wikipedia.org/wiki/Uncanny_valley), so I wouldn't mind a second opinion on this...

wincent:
[...]

Issue #534, opened 2020-03-05
Add release script


izaera:
Like #518 but for real releases (needs more checks and steps to be taken).

izaera:
I have configured liferay-js-publish (#579 and #580) but because we are not yet using the `master` branch it fails. I won't fix it yet and will try to keep going with manual pre-releases until we decide to release the first production release.

jbalsas:
Doing some autumn cleaning as we move this project to https://github.com/liferay/liferay-frontend-projects#js-toolkit

We'll be curating these and creating new ones mostly from scratch to have a better sense of where we want to go.

PR #608, opened 2020-07-01
chore: walk towards release

### LINKED ISSUE #565: Review source maps
Before releasing we must check the generated source maps and their virtual locations so that, when seen in the dev tools, they make sense.

Also, we need to make sure that the source maps are useful and don't have incorrect or too generic mappings.

izaera:
I'm doing some cleanup and refactors in this PR to make the whole project releasable. The idea is to walk towards something that can be released and used in place of bundler 2 whether in portal's build or outside projects.

Given the mixed nature of the commits in this PR and that it touches so many files I recommend to have a look at commit descriptions first and then look at the motivation of the issues contained in the PR.

The rest can be considered cleanup.

Also, this is not final code, but just an improvement over what we have now. Some things still may be refined further but that is not done in this PR to avoid making it more complexand also to avoid drifting too apart from the current state of projects (avoid major breaking changes).

I have tested that this is working creating a React Widget project with bundler 2, then migrating it to 3 (configuration files) and deploying it to portal. However I haven't tested it thoroughly, given that more PRs need to be sent before release. In particular we still need to close #565 which implies a refactoring of the adaptation code. Also, we need to settle on the final configuration syntax (#550) and use `liferay-js-publish` to make releases (#534).


packages/liferay-js-toolkit-scripts/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const {error, print} = format;
+
+export default function main(): void {
+	const scripts = process.argv.slice(2);
+
+	scripts.forEach((script) => {
+		try {
+			// eslint-disable-next-line liferay/no-dynamic-require, @typescript-eslint/no-var-requires
+			const {default: scriptFunction} = require(`./scripts/${script}`);
```
wincent, line 16: FWIW, in liferay-npm-scripts, we effectively hardcode the list of subcommands, which means you don't need a lint suppression and you can benefit from the typechecker (or at least, we would if we used a typechecker in that project...)

The [list of commands](https://github.com/liferay/liferay-npm-tools/blob/3f53aa530ed43d2f3830c66b521177ac071e6ee4/packages/liferay-npm-scripts/src/index.js#L15-L72) isn't so large that the source code becomes unwieldy (and if the list of subcommands grew too long, the tool itself would become unwieldy long before the source did).

packages/liferay-js-toolkit-core/src/project/Probe.ts
```diff
@@ -57,5 +61,19 @@ export default class Probe {
...
+	_hasScriptCalling(tool: string): boolean {
+		const {pkgJson} = this._project;
+		const {scripts} = pkgJson;
+
+		if (!scripts) {
+			return false;
+		}
+
+		return (
+			Object.values(scripts).find((script) => script.startsWith(tool)) !==
```
wincent, line 73: Paranoia: this will conflate `js-toolkit` and `js-toolkit-expansion-pack`/`js-toolkit-yo-momma`/`js-toolkitz` as the same thing. I'd do something like this:

```js
return (
    Object.values(scripts).some((script) => new RegExp(`^\s*${tool}\b`).test(script))
);
```

(Not worrying about escaping `tool` because we have only one caller, internal, which isn't passing anything that would require it; also, using `some()` to avoid comparison against `undefined`.)

packages/liferay-js-toolkit-core/src/project/Project.ts
```diff
@@ -336,15 +356,16 @@ export default class Project {
...
-		this._buildDir = undefined;
 		this._configFile = undefined;
 		this._configuration = undefined;
+		this._outputDir = undefined;
 		this._pkgJson = undefined;
 		this._pkgManager = undefined;
 		this._projectDir = undefined;
 		this._sources = undefined;
-		this._srcDir = undefined;
+		this._sourceDir = undefined;
```
wincent, line 366: noice

wincent:
Looked at every commit. As this is on the 3.x-WIP branch, risk is low — feel free to merge whenever you see fit.

packages/liferay-js-toolkit-scripts/src/index.ts
izaera: Makes sense, since we don't want it to be extensible. I'll change it.

packages/liferay-js-toolkit-core/src/project/Probe.ts
izaera: :+1: Didn't know about boundaries.

packages/liferay-js-toolkit-core/src/project/Project.ts
izaera: #SayNoToAbbreviationsAndUseWholeWordsOnlyEvenIfSentencesAreLongerAndDontFitIn80CharsBecauseWeWantCoherenceInOurCode :muscle:

PR #610, opened 2020-07-02
chore: prepare 2.18.7 release

izaera:
Issue #600 was missing from v2.18.6 because of a mistake made while releasing it, so now I'm sending this PR to release 2.18.7 including it.


izaera:
The builds are green even though it doesn't show up here. Going to merge for release.

PR #611, opened 2020-07-03
refactor: first part of #565 fix

izaera:
This is the first of two PRs trying to make source maps work correctly for bundled projects.

In this first PR I substitute every file transformation operations by calls to the transforms API we started in liferay-js-toolkit-core. This makes all code uniform and especially prevents modifications of JS source code without using an AST and updating the source map accordingly (in theory).

I have tested this PR with a local react project and everything seems to work correctly (even imports because I get React from the portal). However, the main goal of #565 is to make source maps work and they are still broken :-(. 

I need to check why and fix it, but at least, after this PR, every transformation in centralized in one single place where we can transform the source map correctly (when we know how).

I have also made some minor edits to the `liferay-js-toolkit-core` API to prevent naming collisions in the future.


packages/liferay-npm-bundler/src/steps/bundle/plugin/imports-loader.ts
```diff
@@ -3,18 +3,26 @@
...
-import ReportLogger from '../../../report/logger';
+import transformImports from './transformImports';
 
-export default function (content: string): string {
-	const prjDirRelResourceFile = project.dir.relative(this.resourcePath);
+export default function (content: string): void {
+	const callback = this.async();
+
+	// Execute async transformation from webpack async loader
+	asyncTransform(content, this.resourcePath)
```
wincent, line 16: It does seem rather odd to have a function like this unnamed one that calls an async function and neither waits for it, nor provides a means to determine completion (ie. it neither returns the promise nor accepts a callback); that is, it is "fire-and-forget".

packages/liferay-npm-bundler/src/steps/bundle/plugin/transformImports.ts
```diff
@@ -0,0 +1,165 @@
...
+						modified = transformImportDeclaration(
+							log,
+							node,
+							parentNode
+						);
+						break;
+
+					default:
+						break;
+				}
```
wincent, line 40: This is a `switch` which isn't really leveraging any of the strengths of the `switch` construct. ie; plain old `if` would be more succinct, and easier to read, in terms of cognitive load (because you don't have to think about possible fall through, or what the `default` means etc):

```js
if (node.type === 'CallExpression) {
    modified = transformCallExpression(log, node);
} else if (node.type === 'ImportDeclaration') {
    modified = transformImportDeclaration(
        log,
        node,
        parentNode
    );
}
```

packages/liferay-js-toolkit-core/src/transform/json/operation/addDependencies.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Add a list of dependencies to a package.json file
+ *
+ * @param dependencies
+ */
+export default function addDependencies(
+	dependencies: object
+): JsonTransform<PkgJson> {
+	return (async (pkgJson) => {
+		pkgJson.dependencies = pkgJson.dependencies || {};
```
wincent: You don't need this line. The spread of `...pkgJson.dependencies` will work just fine if `pkgJson.dependencies` is `undefined` (`null` would be fine too).

packages/liferay-js-toolkit-core/src/transform/json/operation/deleteDependencies.ts
```diff
@@ -0,0 +1,29 @@
...
+	return (async (pkgJson) => {
+		if (!pkgJson.dependencies) {
+			return pkgJson;
+		}
+
+		packageNames.forEach(
+			(packageName) => delete pkgJson.dependencies[packageName]
+		);
+
+		return pkgJson;
```
wincent: So I gather your goal here is:

1. Do nothing if no `dependencies`.
2. Otherwise, empty of the `dependencies`.

If that's the case, why don't you just do?

```js
if (pkgJson.dependencies) {
    pkgJson.dependencies = {};
}

return pkgJson;
```

In addition to avoid the `forEach` + `delete`, I inverted the sense of the `if` in order to avoid the negation; I think it makes it easier to read.

wincent: Ah, ignore that. I see now that your intent is to delete specific `packageNames` that get passed in from the outside. I mistakenly thought that `packageNames` was just all the packages.

packages/liferay-npm-bundler/src/steps/bundle/plugin/imports-loader.ts
izaera: Tell it to the webpack team :sweat_smile: (see its async loader API :point_right: https://webpack.js.org/api/loaders/#asynchronous-loaders). I will add a commentary to make it more understandable. I agree that, if you haven't heard about the API it looks crazy. That's why I split the module in two methods: one for humans, and the scaffolding :-).

To be fair, webpack's API is probable the only way they could have done it, since it was sync in the beginning and there was no easy way to tell your caller you are async before `Promise`s existed.

packages/liferay-js-toolkit-core/src/transform/json/operation/deleteDependencies.ts
izaera: Yep. I just want to delete a bunch of dependencies, not all :+1:

packages/liferay-js-toolkit-core/src/transform/json/operation/addDependencies.ts
izaera: Quite handy :-)

packages/liferay-npm-bundler/src/steps/bundle/plugin/imports-loader.ts
wincent: Ha, my eyes skipped right past the whole `this.async()` thing.

Jeez.

PR #612, opened 2020-07-07
fix: #565 don't break source maps

izaera:
Don't break source maps for single source transformations.

Note that this doesn't fix the case where one single output file maps to multiple input files.

See https://github.com/liferay/liferay-js-toolkit/issues/565#issuecomment-654883046 for more info.

### DIFF
--- a/packages/liferay-js-toolkit-core/src/index.ts
+++ b/packages/liferay-js-toolkit-core/src/index.ts
@@ -67,7 +67,10 @@ export type {
 export type {default as PkgJson} from './schema/PkgJson';
 
 // JavaScript AST helpers
-export {getProgramStatements as getAstProgramStatements} from './transform/js/ast';
+export {
+	getProgramStatements as getAstProgramStatements,
+	mapNodeLocation as mapAstNodeLocation,
+} from './transform/js/ast';
 export {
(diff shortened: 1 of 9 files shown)


jbalsas:
Like the song... "Don't Go Breaking My SourceMap"...

[Elton John - Don't Go Breaking My Heart (with Kiki Dee)](http://www.youtube.com/watch?v=z0qW9P-uYfM "Don't go breaking my Source Maps")

mfreeman-xtivia:
So will bundler v3 support source maps? for things like React portlets? and if so how do i switch the build mode from production to development since lnbs-build seems to be in charge of that

izaera:
@mfreeman-xtivia we want to support source maps, but it's proving much more difficult than we thought at first, second and third :-(. Basically, we need to do something that no one else has done yet, AFAIK, which is translating source maps for multiple file bundles.

wincent:
Going to close this as we've migrated everything in the toolkit to the [liferay-frontend-projects monorepo](https://github.com/liferay/liferay-frontend-projects). We can re-send the PR there if we want to revisit this line of development in the future. This (liferay-js-toolkit) repo will soon be moved into archived (read-only) mode.

Issue #613, opened 2020-07-09
blank after "Bundle-SymbolicName" in META-INF/MANIFEST.MF cause jar failed to be deployed


Wujae:
The blank after property "Bundle-SymbolicName", "Bundle-Version", "Bundle-Name", "Web-ContextPath" in META-INF/MANIFEST.MF generated by command "liferay-npm-bundler“, seems to cause OSGiException while deploying the jar.

I traced the source code and found out that :
**get content()** method in **liferay-npm-bundler/src/jar/manifest.ts** return serveral blanks after those property mentioned above.

Wujae:
I forgot to add "description" proprety in package.json which cause a blank value in "Bundle-Name"

izaera:
I will reopen it then, so that we can use the `name` when no `description` is found. :+1:

PR #614, opened 2020-07-10
fix: #613 don't add empty lines to MANIFEST.MF

izaera:
Manually tested, but I have also added unit tests (and found an error!)

### DIFF
--- a/packages/liferay-npm-bundler/src/jar/__tests__/manifest.test.js
+++ b/packages/liferay-npm-bundler/src/jar/__tests__/manifest.test.js
@@ -16,7 +16,7 @@ Tool: liferay-npm-bundler-2.18.7
 	);
 });
 
-it('set bundleSymbolicName works', () => {
+it('includes Bundle-SymbolicName when provided', () => {
 	const manifest = new Manifest();
 
 	manifest.bundleSymbolicName = 'My wonderful bundle';
@@ -24,13 +24,13 @@ it('set bundleSymbolicName works', () => {
 	expect(manifest.content).toEqual(
 		`Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
-Tool: liferay-npm-bundler-2.18.7
 Bundle-SymbolicName: My wonderful bundle
+Tool: liferay-npm-bundler-2.18.7
 `
 	);
 });
 
-it('set bundleVersion works', () => {
+it('includes Bundle-Version when provided', () => {
 	const manifest = new Manifest();
 
 	manifest.bundleVersion = '1.0.0';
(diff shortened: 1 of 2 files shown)


packages/liferay-npm-bundler/src/jar/__tests__/manifest.test.js
```diff
@@ -0,0 +1,136 @@
...
+
+	expect(manifest.content).toEqual(
+		`Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Tool: liferay-npm-bundler-2.18.7
+`
+	);
+});
+
+it('set bundleSymbolicName works', () => {
```
wincent, line 19: The idea of `it` is that it should read like a sentence with what follows, so all these:

```js
it('set bundleSymbolicName works', () => {});
```

don't really satisfy that. Also "works" is a weak specification of behavior because it doesn't encode what "working" behavior actually is. So I'd suggest something like:

```js
it('includes bundleSymbolicName', () => {});
```

or:

```js
it('includes bundleSymbolicName when provided', () => {});
```

(There's also `test`, which is an alias for `it`, but we don't use that anywhere. It lets you write stuff like `test('set bundleSymbolicName works', () => {})`, but it's still desirable to use a better word than "works".)

packages/liferay-npm-bundler/src/jar/manifest.ts
```diff
@@ -81,15 +81,29 @@ export default class Manifest {
...
-${header('Bundle-Version', this._bundleVersion)} 
-${header('Bundle-Name', this._bundleName)} 
-${header('Web-ContextPath', this._webContextPath)} 
-${capabilities('Provide-Capability', this._provideCapabilities)}
-${capabilities('Require-Capability', this._requireCapabilities)}`;
+`;
+
+		content += header('Bundle-SymbolicName', this._bundleSymbolicName);
+		content += header('Bundle-Version', this._bundleVersion);
+		content += header('Bundle-Name', this._bundleName);
```
wincent, line 91: Is it OCD to want these ordered?

izaera: Yes. But I will order it following the same criterion we use in portal to avoid disturbances in the space-time continuum.

izaera: We don't want a cataclysmic multiversal implosion, do we?

wincent: Yeah, not ready for that yet...

izaera: It turns out that it is alphabetically ordered in portal

unexpected

Issue #620, opened 2020-08-03
add <!DOCTYPE html> to .webpack > index.html


alimahsis:
Without **`<!DOCTYPE html>`** some packages not work properly. For example, `file-pond`, `beautiful-dnd` and `katex` packages have problems. For example, in the `katex` package this error occurs:

`Warning: KaTeX doesn't work in quirks mode. Make sure your website has a suitable doctype.`

Must be added to the **.webpack > index.html** file.

wincent:
Going to do a mass-transfer of issues (including this one):

- **From:** https://github.com/liferay/liferay-js-toolkit/issues
- **To:** https://github.com/liferay/liferay-frontend-projects/issues

Note that many of the issues that we move may be stale, but we can close those once we've had time to review them. In the meantime, doing a bulk move makes is cheap and it will allow us to mark the liferay-js-toolkit repo as archived (read-only), which is an important final step in the consolidation process in order to avoid confusion.

PR #621, opened 2020-08-04
fix: add DOCTYPE to disable Quirk Mode (#620)

alphazed110:
#620 
Please merge.
thanks.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -10,5 +10,5 @@
 		"packages/*"
 	],
 	"useWorkspaces": true,
-	"version": "2.18.7"
+	"version": "2.18.8"
 }
--- a/packages/babel-plugin-add-module-metadata/package.json
+++ b/packages/babel-plugin-add-module-metadata/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "babel-plugin-add-module-metadata",
-	"version": "2.18.7",
+	"version": "2.18.8",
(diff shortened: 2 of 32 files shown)


wincent:
This looks harmless enough to me, but @izaera can take a look and merge when he gets back. We've been following [the Conventional Commits spec](https://www.conventionalcommits.org/en/v1.0.0/) in this repo for a while now though, so both the commit and the PR title should be edited to start with the prefix "fix:".

Seeing as fixing the commit title requires an `--amend`, you may as well [provide a description](https://github.com/liferay/liferay-frontend-guidelines/blob/master/general/commit_messages.md) in the commit message body too. The issue (#620) contains useful context, but in the commit message all that is reduced to just "fix-620 add DOCTYPE".

alphazed110 (commented):
review done.

wincent:
Going to close this as we've migrated everything in the toolkit to the [liferay-frontend-projects monorepo](https://github.com/liferay/liferay-frontend-projects). Please re-open it there if you consider this to be something that is still useful. Concretely, note that the project files for v2 of the toolkit have moved into the [`maintenance/projects/js-toolkit/` directory](https://github.com/liferay/liferay-frontend-projects/tree/master/maintenance/projects/js-toolkit).

This (liferay-js-toolkit) repo will soon be moved into archived (read-only) mode, which means we won't be able to make any more updates or edits here.

Issue #625, opened 2020-08-13
Add specific build for fragments


izaera:
With big probability we may need to tailor a specific build for fragments as part of [LPS-102280](https://issues.liferay.com/browse/LPS-102280). We could also do it in the fragments toolkit with webpack, but I believe it will be much better to do it in the JS Toolkit to reuse the functionality we already have in place to transform files and projects.

jbalsas:
Closing since I think this is done now, or should be moved to https://github.com/liferay/liferay-frontend-projects#js-toolkit

PR #627, opened 2020-08-18
feat: #625 add specific build for fragments

izaera:
This is a revamp of #626 applying a different approach for the server side.

I think the PR is self evident by reading the comments. Basically what we 
want to do is generating files for the fragments CLI to use as input.

This means compiling each React based fragment as a single webpack bundle
that imports `react` from the Portal and bundles any other dependency inside
the webpack bundle.

These bundles are then loaded by the fragment renderer in the Portal, 
leveraging ReactRenderer service.

You can see an example of how a fragment project would look like when using
this PR in [this repo](https://github.com/izaera/react-fragment/tree/5bc7dbee5a4c79459edcdd2f609b794bd1848ec3).

The server counterpart of this PR is [this other](https://github.com/liferay-frontend/liferay-portal/pull/268).


jbalsas:
Thanks @izaera! Can you point out what's the counterpart PR in DXP?

Also... is all the code here final, or still somehow PoC material? How will this be consumed from the fragment-cli?

izaera:
Note that this solution makes it impossible to import any package other than `react` from the Portal (as noted [here](https://github.com/liferay/liferay-js-toolkit/pull/626#issuecomment-675466711)) and [here](https://github.com/liferay/liferay-js-toolkit/pull/627/commits/3dd06f20149b15e146e94304221a60c94922a8da).

This is because fragments will be dynamic `JSModule`s for the `NPMRegistry`, but they will live inside `fragment-impl` `JSBundle`, so they cannot resolve any package they want, only the ones we already provision in `fragment-impl`.

If, in the future, we want to allow importing any random package, we need to attach more information to the fragments (namely the `package.json`) and make each fragment a full dynamic `JSBundle` perhaps... That is left for the next iteration in case we want to do it.

izaera:
> Also... is all the code here final, or still somehow PoC material? How will this be consumed from the fragment-cli?

The code can be considered final, but we need to change the fragments CLI to call the bundler if we want to put all source files in "src" instead of "react-src" (because the fragments CLI hardcodes the "src" directory and is not prepared to cooperate with any other build tool now).

Or, if we don't want the fragments CLI to call the bundler, at least make its src folder configurable (similar to what we do in portal for the gradle/bundler integration).

izaera:
For the record, this is how a fragment bundle would look like :point_right: https://gist.github.com/izaera/615a1156ade8c051d5b8945d45079e87

izaera:
@jbalsas one thing we don't cover yet is how to define [the exports](https://github.com/izaera/react-fragment/blob/5bc7dbee5a4c79459edcdd2f609b794bd1848ec3/liferay-npm-bundler.config.js#L9) for fragment projects. We may leave it to:

1) the user (by hand)
2) the fragments CLI before calling the bundler
3) make the bundler autodetect them scanning the project
4) other solution...

izaera:
We still need to define how we are going to inform the server that a fragment is of React type. Currently we are detecting it because it starts with `Liferay.Loader.define` but that's clearly not a good solution. See [this change](https://github.com/liferay-frontend/liferay-portal/pull/268/commits/9aa2b1c5b4b64c3374b41cc778e1fbef1252c85e#diff-3e86fcbeff866b4f57dfc34548995598R204).

packages/liferay-js-toolkit-core/src/project/Probe.ts
```diff
@@ -10,6 +10,7 @@ export enum ProjectType {
 	CREATE_REACT_APP = 'create-react-app',
 	ANGULAR_CLI = 'angular-cli',
 	VUE_CLI = 'vue-cli',
+	FRAGMENT = 'liferay-fragment',
```
wincent, line 13: Plz sort.

packages/liferay-npm-bundler/src/steps/adapt/fragment/index.ts
```diff
@@ -0,0 +1,75 @@
...
+
+		fs.ensureDirSync(destFile.dirname().asNative);
+
+		fs.copyFileSync(
+			project.sourceDir.join(file).asNative,
+			destFile.asNative
+		);
+	});
+
+	log.info(`Copied ${files.length} static files to output directory`);
```
wincent, line 74: `<ocd>`

One way to avoid a potential bothersome:

> Copied 1 static files to output directory

without special logic is to write messages like this:

> Static files copied: 1

`</ocd>`

packages/liferay-js-toolkit-core/src/project/Probe.ts
izaera: You are never satisfied, are you? 

:joy:

packages/liferay-npm-bundler/src/steps/adapt/fragment/index.ts
izaera: I always forget that trick :-(

izaera:
I'm resending a new version of this in a minute.

PR #633, opened 2020-09-01
chore: create CONTRIBUTING.md file

izaera:

PR #634, opened 2020-09-01
docs: update CHANGELOG.md

wincent:
This changelog was originally managed by [gitlab-changelog-generator](https://github.com/skywinder/gitlab-changelog-generator), but after v1.6.1 we started using GitHub milestones instead to record changes for each release, so we stopped updating it.

On the 3.x-WIP branch we have started using liferay-changelog-generator and our new CONTRIBUTING guidelines, adding in [this PR](https://github.com/liferay/liferay-js-toolkit/pull/633), indicate that we should do the same on this branch too.

So, I used the following command to generate the missing range:

    npx liferay-changelog-generator --from=v1.6.1 --regenerate --version=v2.19.0

And then edited by hand to note that most of the details for those releases are still to be found in the milestones (liferay-changelog-generator bases its output on the merge commit messages, which mostly aren't present in that range).

Finally, I stripped off the last version (v2.19.0), because I'll generate that when I actually cut that release (ie. after this PR).

Subsequent updates should be easy, just like the CONTRIBUTING says:

    npx liferay-changelog-generator --version=$VERSION


wincent:
Assuming this is non-controversial, going to merge it.

PR #636, opened 2020-09-01
fix: make namespace-amd-define work with css-loader

wincent:
As noted in [this liferay-portal PR](https://github.com/liferay-frontend/liferay-portal/pull/345#issuecomment-684867451), code like this which is produced by the css-loader:

```js
function defineModule() {
  define("frontend-js-web@4.0.27/liferay/modal/Modal.scss", ['module', 'exports', 'require'], function (module, exports, require) {
    var define = undefined;
    var global = window;
    {
      module.exports = link;
    }
  });
}
```

winds up with a naked `define` in it, instead of the expected `Liferay.Loader.define`.

This behavior regressed in 0c14c6970503d66d73cb5b311d15ce5748d9eca5.

Now, I don't have enough context to know whether this is the right fix, because the issues listed in that commit are kind of opaque to me. Hopefully @izaera can confirm or refute that the proposed change here is safe.

Of course, it's not very clever; if you have code like this:

```js
function a() {
  define([], function () {});
}

define([], function () {});
```

It is going to namespace the first `define` it sees and warn about the second one, but it's totally arbitrary; there is nothing "special" about the first one that appeared higher up in the file. If move the function to the bottom, then the other `define` "wins" and gets transformed.

### DIFF
--- a/packages/babel-plugin-namespace-amd-define/src/__tests__/__snapshots__/index.test.js.snap
+++ b/packages/babel-plugin-namespace-amd-define/src/__tests__/__snapshots__/index.test.js.snap
@@ -66,3 +66,10 @@ exports[`namespaces unqualified define calls 1`] = `
 "
 Liferay.Loader.define([], function () {});"
 `;
+
+exports[`namespaces unqualified define calls that are wrapped 1`] = `
+"
+function defineModule() {
+	Liferay.Loader.define([], function () {});
+}"
+`;
--- a/packages/babel-plugin-namespace-amd-define/src/__tests__/index.test.js
+++ b/packages/babel-plugin-namespace-amd-define/src/__tests__/index.test.js
@@ -61,6 +61,22 @@ it('namespaces unqualified define calls', () => {
 	expect(code).toMatchSnapshot();
 });
 
+it('namespaces unqualified define calls that are wrapped', () => {
+	// eg. the css-loader produces wrappers like this.
+	const source = `
+		function defineModule() {
+			define([], function () {});
+		}
+	`;
+
+	const {code} = babel.transform(source, {
+		filename: __filename,
+		plugins: [plugin],
+	});
+
+	expect(code).toMatchSnapshot();
+});
+
 it('does not namespace already qualified define calls', () => {
 	const source = `
 	Other.Namespace.define([], function(){})
--- a/packages/babel-plugin-namespace-amd-define/src/index.js
+++ b/packages/babel-plugin-namespace-amd-define/src/index.js
@@ -37,7 +37,11 @@ export default function() {
 				}
 
 				if (!firstDefineNamespaced) {
-					if (path.scope.parent === null) {
+					if (
+						path.scope.parent === null ||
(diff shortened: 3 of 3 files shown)


wincent:
Just gonna rebase this to get the test fixes from `master` that are making this red instead of green.

wincent:
Testing this manually in liferay-portal, I think there is something additional going on. While this change obviously fixes the test case included in this commit, there is something happening in the liferay-portal build that is causing this line to have no effect (ie. I can `console.log` it and see it supposedly working, but the transformed source doesn't show up on the disk):

```js
path.node.name = `${namespace}.define`;
```

Some interaction between the various plugins, or something. Going to continue looking into that.

wincent:
Ah no, I lie. It is working. I was just cheating in order to test the changes locally, running `../../../node_modules/.bin/liferay-npm-scripts build` and then inspecting the wrong place (`classes` instead of `build`). If I look in `build`, the changes are there, as expected.

phew

wincent:
See https://github.com/liferay/liferay-js-toolkit/pull/632#issuecomment-684949352 for screenshots of this working in liferay-portal now.

wincent:
This is green:

- https://travis-ci.org/github/liferay/liferay-js-toolkit/builds/723115841
- https://travis-ci.org/github/liferay/liferay-js-toolkit/builds/723115843

Even though GitHub is showing it as "Pending".

izaera:
Given that [the commit that caused this regression](https://github.com/liferay/liferay-js-toolkit/commit/0c14c6970503d66d73cb5b311d15ce5748d9eca5) was introduced to fix #589 and it seems we don't have tests to explicitly test the condition in #589, I will add a new test for that and, if it passes, I guess we may merge this.

Another possible solution to this _namespacing nightmare_ could be dropping the namespace plugin and doing the wrapping and namespacing in one shot (we already do that in bundler 3). What do you think of this solution @wincent ?

wincent:
> Another possible solution to this _namespacing nightmare_ could be dropping the namespace plugin and doing the wrapping and namespacing in one shot (we already do that in bundler 3). What do you think of this solution @wincent ?

I am not qualified to know the answer to that one. In the abstract though, my preference is for the smallest possible change (which probably means _not_ factoring away the plugin), because in theory v2 is in "maintenance" mode and the less we touch it the better. On to the future, v3.

izaera:
I'll add the test then and continue. If this issue reappears in the future we may try the other approach.

izaera:
Looks like we [already have the test in place](https://github.com/liferay/liferay-js-toolkit/blob/07e6565bafad3f2a0aba63cea8872da03f157c13/packages/babel-plugin-namespace-amd-define/src/__tests__/index.test.js#L78), so going to merge this.

wincent:
> Looks like we [already have the test in place](https://github.com/liferay/liferay-js-toolkit/blob/07e6565bafad3f2a0aba63cea8872da03f157c13/packages/babel-plugin-namespace-amd-define/src/__tests__/index.test.js#L78), so going to merge this.

celebrate

PR #642, opened 2020-09-16
chore: enhance QA scripts

izaera:
Two simple changes to make QA scripts more robust. Functionality remains the same.


izaera:
Going to merge this, as QA may need it and doesn't affect production code.

Issue #631, opened 2020-08-28
Make css-loader use "preload" when available


wincent:
In service of [LPS-115876](https://issues.liferay.com/browse/LPS-115876), the idea is to have the css-loader emit:

```html
<link as="style" rel="preload" />
```

tags instead of:

```html
<link rel="stylesheet" />
```

ones.

As noted [here](https://www.smashingmagazine.com/2016/02/preload-what-is-it-good-for/):

> On top of that, preload **does not block the window’s `onload` event**, unless the resource is also requested by a resource that blocks that event.

In practice, we don't expect this to make page loads perceptibly faster, but it should at least avoid Lighthouse reporting onload events as delayed.

Related community Slack discussion: https://liferay-community.slack.com/archives/CB2QQ3M5F/p1598598237007500

PR #645, opened 2020-09-17
revert: feat(css-loader): use preload when available (#631)

wincent:
This reverts commit 1423ca31e3a1f10, because the feature doesn't work as expected, as [described in this comment](https://github.com/liferay-frontend/liferay-portal/pull/345#issuecomment-694221607).

In short, our testing showed the loader producing code that correctly inserted `preload` links in the `head`, and the network pane showed the corresponding resources being fetched, and in the right order. But the associated styles were not being applied to the DOM, leading to visual inconsistencies, as well as undesirable console spew of the form:

> The resource .../css/ApplicationsMenu.css was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.

To fix this, we might be able to add a second, non-`preload`-ing `link` tag, but that would defeat the purpose of our original change, which was to improve our Lighthouse metrics.

So, reverting this for now. We might revisit `preload` for this or some other purpose in the future.

Original PR: https://github.com/liferay/liferay-js-toolkit/pull/632

### DIFF
--- a/packages/liferay-npm-bundler-loader-css-loader/src/__tests__/__snapshots__/index.test.js.snap
+++ b/packages/liferay-npm-bundler-loader-css-loader/src/__tests__/__snapshots__/index.test.js.snap
@@ -3,24 +3,7 @@
 exports[`java projects correctly generates JS module 1`] = `
 "
 var link = document.createElement(\\"link\\");
-
-var preload;
-
-try {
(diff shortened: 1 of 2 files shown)


wincent:
Simple revert, and I have to release 821 packages before I can update liferay-portal, so I'm going to merge this.

PR #648, opened 2020-09-22
feat: #625 add specific build for fragments

izaera:
This is the final PR for the fragments build support. It supersedes #627.

I think the PR is self evident by reading the commits. Basically what we
want to do is generating files for the fragments CLI to use as input.

This means compiling each React based fragment as a single webpack bundle
that imports react from the Portal and bundles any other dependency inside
the webpack bundle.

These bundles are then loaded by the fragment renderer in the Portal,
leveraging ReactRenderer service.

You can see an example of how a fragment project would look like when using
this PR in this [repo](https://github.com/izaera/react-fragment/tree/a589b99da016c0c6e4a8425d20ce7b6c0f180fa2).


packages/liferay-npm-bundler/src/util/abort.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * SPDX-FileCopyrightText: © 2020 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+/**XX
+ * SPDX-FileCopyrightText: © 2017 Liferay, Inc. <https://liferay.com>
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
```
wincent, line 9: What happened here?

wincent:
I only skimmed this, but as it is isolated to `3.x-WIP` it seems like the risk is pretty low, so from my side this is a green light...

green-light

packages/liferay-npm-bundler/src/util/abort.ts
izaera: OMG :scream:

izaera:
@wincent also, it's almost the same as the previous PR (except for the refactors) and it was already reviewed.

PR #653, opened 2020-09-29
feat: remove copy assets step for fragments

izaera:
This step was a temporary workaround to integrate with the Fragments CLI. Now
that the Fragments CLI's build can be properly extended, we don't need it any
more because the Fragments CLI itself will do it.


izaera:
I'm going to merge this once it is green because it is blocking the Fragments CLI team and it's a change that cannot be really tested until we integrate it into the Fragments CLI build.

wincent (approved):
Looks fine to me.

PR #656, opened 2020-09-30
fix: use a module name placeholder for fragments

izaera:
Because the server gives fragment modules a name on-the-fly (based on the fragment entry id) we cannot write any name during build time, so instead we write a placeholder (`__FRAGMENT_MODULE_NAME__`) that the server replaces by the real name when registering the AMD module in the `NPMRegistry`.

This have been tested in my laptop but it needs a very special setup and modifications to the Fragments CLI, so we still need to integrate everything together.

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,1 +1,2 @@
+--publish.tag next
 --version.preid pre
--- a/packages/liferay-npm-bundler/src/adapt/liferay-fragment/index.ts
+++ b/packages/liferay-npm-bundler/src/adapt/liferay-fragment/index.ts
@@ -37,12 +37,10 @@ async function transformBundles(): Promise<void> {
 
 		const destFile = project.outputDir.join(moduleName);
 
-		const {name, version} = project.pkgJson;
-
 		await transformJsSourceFile(
 			sourceFile,
 			destFile,
-			wrapModule(`${name}@${version}/${moduleName}`, {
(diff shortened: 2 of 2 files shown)


.yarnrc
```diff
@@ -1 +1,2 @@
+--publish.tag next
```
wincent, line 1: 👍

packages/liferay-npm-bundler/src/adapt/liferay-fragment/index.ts
```diff
@@ -37,12 +37,10 @@ async function transformBundles(): Promise<void> {
 
 		const destFile = project.outputDir.join(moduleName);
 
-		const {name, version} = project.pkgJson;
-
 		await transformJsSourceFile(
 			sourceFile,
 			destFile,
-			wrapModule(`${name}@${version}/${moduleName}`, {
+			wrapModule(`__FRAGMENT_MODULE_NAME__`, {
```
wincent, line 43: Should probably be single quotes (don't know if Prettier or ESLint will complain).

wincent (approved):

packages/liferay-npm-bundler/src/adapt/liferay-fragment/index.ts
izaera: They don't complain, but I'll change to single since I've removed any interpolation :+1:

